PR #2, opened 2018-10-15
Fixed checkbox style

alansouzati:


index.js
```diff
@@ -82,6 +86,18 @@ export const hpe = deepFreeze({
     },
     extend: 'letter-spacing: 0.04167em;',
   },
+  checkBox: {
+    icon: {
+      extend: css`
```
ericsoderberghp: I saw this in the `grommet` theme too. Why can't we put this in `base`?

alansouzati: because growing the theme to take the full width of the box seems specific to us. Our theme does not want that. it is centered, as most checkboxes. it was hard to us to customize it with this there.

.eslintrc
```diff
@@ -0,0 +1,40 @@
+{
+  "parser": "babel-eslint",
+  "extends": "airbnb",
+  "env": {
+    "node": true,
+    "es6": true
+  },
+  "globals": {
```
ericsoderberghp: Can we remove all of these?

.eslintrc
```diff
@@ -0,0 +1,40 @@
...
+    "contextTypes": true,
+    "state": true,
+    "Event": true,
+    "fail": true,
+    "afterEach": true,
+    "beforeEach": true,
+    "MouseEvent": true,
+    "requestAnimationFrame": true
+  },
+  "rules": {
```
ericsoderberghp: Can we remove most of these?

__tests__/index-test.js
```diff
@@ -0,0 +1,5 @@
+import { hpe } from '../src/js';
+
+test('Loads hpe theme', () => {
+  expect(hpe).toMatchSnapshot();
```
ericsoderberghp: Not much of a test ;)

.eslintrc
alansouzati: I think so

.eslintrc
alansouzati: i think so

PR #10, opened 2020-02-03
Update new HPE theme and brand

ShimiSun:


ericsoderberghp:
I think we'll need the version number changed in package.json.

Also, can we get the `formField` theme updates in as well?

ShimiSun:
All up to date @ericsoderberghp

ShimiSun:
On the last commit, I took the complete theme from the Designer HPE theme and replaced the current theme, I had to clean it a lot so styles that already exist on `base.js` won't be redundant.
After the clean-up, I've applied the missing style pieces from the aries theme.
Currently, the components that are missing from the designer styling has a comment for easier review.

src/js/index.js
```diff
@@ -9,92 +11,442 @@ const deepFreeze = obj => {
...
-
         @font-face {
           font-family: "Metric";
           src: url("https://hpefonts.s3.amazonaws.com/web/MetricHPE-Web-Bold.woff") format('woff');
           font-weight: 700;
         }
-
         @font-face {
-          font-family: "Metric";
+          font-family: "Metric";
```
ShimiSun:
```suggestion
          font-family: "Metric";
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "grommet-theme-hpe",
-  "version": "0.4.2",
+  "version": "2.0.0",
```
ShimiSun:
```suggestion
  "version": "1.0.0",
```

src/js/index.js
```diff
@@ -200,6 +201,20 @@ export const hpe = deepFreeze({
...
+  formField: {
+    border: {
+      side: 'all',
+    },
+    content: {
+      pad: {
+        horizontal: 'small',
+        bottom: 'small',
+      },
+      disabled: {
```
ericsoderberghp: `disabled` should be outside of `content`.

ShimiSun: I think you commented on an old commit.
I actually took `disabled` out to base.js, so it shouldn't appear on this theme.
@ericsoderberghp

ShimiSun: See https://github.com/grommet/grommet/pull/3722/files line 551 on the edited code.

src/js/index.js
```diff
@@ -9,92 +11,437 @@ const deepFreeze = obj => {
...
-  });
-
-colorArray(accentColors, 'accent');
-colorArray(neutralColors, 'neutral');
-Object.keys(statusColors).forEach(color => {
-  colors[`status-${color}`] = statusColors[color];
-});
-
 export const hpe = deepFreeze({
+  // name: 'HPE 2',
```
ericsoderberghp: Why are we commenting these out?

src/js/index.js
```diff
@@ -9,92 +11,437 @@ const deepFreeze = obj => {
...
-    analog: {
-      second: {
-        color: {
-          dark: '#01a982',
-          light: '#01a982',
+  heading: {
+    level: {
+      '1': {
+        small: {
+          size: '43.6px',
```
ericsoderberghp: So, I would think all of these sizes would be the same as base.js, since grommet also uses 24px base spacing?

src/js/index.js
```diff
@@ -9,92 +11,437 @@ const deepFreeze = obj => {
...
+        light: '#000000',
+      },
+      'text-weak': {
+        dark: '#CCCCCC',
+        light: '#444444',
+      },
+      'text-xweak': {
+        dark: '#999999',
+        light: '#666666',
+      },
```
ShimiSun: Could be partially removed/cleaned-up after merging https://github.com/grommet/grommet/pull/3722

src/js/index.js
ShimiSun: I commented-out attributes that don't have actual impact on the code. So we wouldn't imply this attribute has another meaning of just the name.

ericsoderberghp: It means the HPE theme will be inconsistent with themes coming from the theme designer then, anyone who builds a theme with the theme designer will have these properties set. And, I don't think that's a problem, it allows the theme to be imported/pasted into the theme designer and then modified.

src/js/index.js
ShimiSun: Partly accurate, the `baseSpacing` is the same, but these values are caculated according to `fontSizing`, and the fontSizing of the HPE theme is different than the base.js.
e.g.
Screen Shot 2020-02-04 at 4 44 05 PM

ericsoderberghp: Ah, OK.

src/js/index.js
ShimiSun: we agreed to keep it all.

ShimiSun:
The theme had been tested on aries-site and works as expected.
I've created a stable branch to be able to test the theme via grommet storybook, grommet-site and other HPE apps.
I'll be merging this PR to be able to move on to the next step, so this theme content will be available on stable.

PR #11, opened 2020-02-06
Publish grommet-theme-hpe to a stable branch

ShimiSun:
Add a script to create a stable branch, so users can point to the master branch of grommet-theme-hpe for preliminary testing of the theme, before the official publishing.

### DIFF
--- a/tools/release-stable.js
+++ b/tools/release-stable.js
@@ -1,0 +1,29 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable no-console */
+import del from 'del';
+import fs from 'fs-extra';
+import git from 'simple-git/promise';
+import path from 'path';
+
+const repoURL = `https://${process.env.GH_TOKEN}@github.com/grommet/grommet-theme-hpe.git`;
+const localFolder = path.resolve('.tmp/grommet-theme-hpe');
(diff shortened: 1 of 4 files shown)


ShimiSun:
I'm merging so I can test the functionality is actually working and the release-stable is being called as expected. Feel free to still drop review comments, although the PR is merged.

ShimiSun:
Yep, we got it!
https://github.com/grommet/grommet-theme-hpe/tree/stable

image

tools/release-stable.js
```diff
@@ -0,0 +1,29 @@
...
+import del from 'del';
+import fs from 'fs-extra';
+import git from 'simple-git/promise';
+import path from 'path';
+
+const repoURL = `https://${process.env.GH_TOKEN}@github.com/grommet/grommet-theme-hpe.git`;
+const localFolder = path.resolve('.tmp/grommet-theme-hpe');
+const localDist = path.resolve('dist');
+
+if (process.env.CI) {
```
halocline: Small nit. I believe all environment variables evaluate as strings, so this block will always execute regardless of the value assigned to `CI` (e.g. if CI=false, this block will still execute). Suggestion: `if (process.env.CI === 'true')`

PR #16, opened 2020-03-13
Update from design-system as of 3/12/20

taysea:
This moves any theme updates from the design-system repo as of this week's sprint into the published theme.

The only remaining pieces in the design-system aries.js file are extend functions that utilize normalizeColor function from grommet/utils. @halocline mentioned wanting to think more on if we can implement the extend styling without these functions before thinking about adding grommet as a peer dependency. I have left those aspects of the theme in the design-system theme file for now.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -101,8 +101,11 @@ export const hpe = deepFreeze({
       'graph-0': 'orange',
       'graph-1': 'blue',
       'graph-2': 'purple',
-      focus: 'green',
+      focus: 'teal!',
       placeholder: 'text-weak',
+      input: {
+        weight: 500,
+      },
     },
     font: {
       family: "'Metric', Arial, sans-serif",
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -225,11 +228,47 @@ export const hpe = deepFreeze({
   checkBox: {
     gap: 'medium',
     color: 'text-strong',
+    toggle: {
+      color: {
+        dark: global.colors['text-strong'].dark,
+        light: global.colors['text-strong'].dark,
```
ShimiSun: should both dark+light use the dark color? it doesn't feel right.

taysea: This matches what is in the sticker sheet -- it means that the toggle knob (when not checked) is always #FFFFFF.

taysea: Resolved via slack -- knob should use background/background front color

src/js/index.js
```diff
@@ -101,8 +101,11 @@ export const hpe = deepFreeze({
       'graph-0': 'orange',
       'graph-1': 'blue',
       'graph-2': 'purple',
-      focus: 'green',
+      focus: 'teal!',
       placeholder: 'text-weak',
+      input: {
+        weight: 500,
```
ShimiSun:
```suggestion
        weight: 500, // normal 
```

ShimiSun:
```suggestion
        weight: 500, // normal 
```

PR #21, opened 2020-03-20
Changed active-background colors

ShimiSun:


src/js/index.js
```diff
@@ -54,7 +54,10 @@ export const hpe = deepFreeze({
         dark: '#444444',
       },
       control: 'brand',
-      'active-background': 'background-contrast',
+      'active-background': {
+        dark: '#FFFFFF1F',
+        light: '#CCCCCC99',
```
taysea: We want 99? The design has 60%

ShimiSun: 60% == 99. 
99 doesn't mean 99% cause letters are also part of the calculation.
Take a look at this resource:
https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4

taysea: Oh nice! I didn't know that.

gregfuruiye: Yes, so the hexidecimal code for the transparency is different from the actual value we are using in figma. I'm using this resource to specify the transparency for hex colors: https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4

gregfuruiye: When we use transparency in our color styles for the design system, we can work to provide the hex number to include the transparency value as well.

Issue #32, opened 2020-04-01
Problems with graph colors against background-front in dark mode


ecircenis:
The graph-2 color is almost undistinguishable from background-front in dark mode.
The graph-0 and graph-4 colors in dark mode are nearly identical to each other.
The graph-1 color has low contrast against background-front.

![image](https://user-images.githubusercontent.com/17883783/78188657-db4f9e80-742d-11ea-8605-6095111afb90.png)

In general, in dark mode, the graph colors should be lighter because they will typically be shown against a background-front color.

ShimiSun:
I think the issue is resulted from not using the 'bang' colors.
So in this case if we'll use
`'graph-0': 'orange!',` instead of `'graph-0': 'orange',` we shouldn't have a problem.

https://github.com/grommet/grommet-theme-hpe/blob/6ff15cc51728419e0f7cce4cf4876f0da797e6c2/src/js/index.js#L113

ShimiSun:
I've filed a PR to address this issue.

ShimiSun:
This issue should be fixed with my recent PR, hence I'll be closing it. Thanks for reporting.

PR #34, opened 2020-04-01
Changed graph-* colors

ShimiSun:
Following the issue https://github.com/grommet/grommet-theme-hpe/issues/32
on bad contrast in dark mode, I fixed the existing colors.

I've also added undefined for graph colors since we shouldn't use the grommet colors, but a better approach (backward compatibility one) would be let hpe theme have its own graph colors (3 & 4).

Thoughts?


src/js/index.js
```diff
@@ -110,9 +110,11 @@ export const hpe = deepFreeze({
         light: '#FFEB59',
       },
       'yellow!': '#FEC901',
-      'graph-0': 'orange',
-      'graph-1': 'blue',
-      'graph-2': 'purple',
+      'graph-0': 'orange!',
+      'graph-1': 'blue!',
+      'graph-2': 'purple!',
+      'graph-3': undefined,
```
ShimiSun, line 116:
```suggestion
      'graph-3': 'yellow!',
```

src/js/index.js
```diff
@@ -110,9 +110,11 @@ export const hpe = deepFreeze({
...
+      'graph-4': undefined,
```
ShimiSun, line 117:
```suggestion
      'graph-4': 'teal!',
```

PR #36, opened 2020-04-02
Update background contrast

gregfuruiye:
small tweaks than include appropriate transparency

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -38,8 +38,8 @@ export const hpe = deepFreeze({
         light: '#FFFFFF',
       },
       'background-contrast': {
-        dark: '#FFFFFF12',
-        light: '#11111108',
+        dark: '#1FFFFFFF',
+        light: '#99CCCCCC',
       },
       icon: 'text',
       text: {


src/js/index.js
```diff
@@ -38,8 +38,8 @@ export const hpe = deepFreeze({
         light: '#FFFFFF',
       },
       'background-contrast': {
-        dark: '#FFFFFF12',
-        light: '#11111108',
+        dark: '#1FFFFFFF',
+        light: '#99CCCCCC',
```
ericsoderberghp, line 42: This looks like we've swapped the opacity to the first bytes?

gregfuruiye: @ericsoderberghp I was following this convention:  https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4#how-to-set-transparency-with-hex-value-

ericsoderberghp: I believe that's only for Android. Here's the way grommet does it for regular browsers: https://css-tricks.com/8-digit-hex-codes/

ShimiSun: Thanks for catching that @ericsoderberghp

PR #39, opened 2020-04-03
Update colors

gregfuruiye:
Color updates as related to the Text Color Styles:
https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/hpe-color?node-id=0%3A1


ericsoderberghp:
This needs to be merged with master to pick up the graph-* color changes Shimi did.

src/js/index.js
```diff
@@ -43,20 +43,20 @@ export const hpe = deepFreeze({
       },
       icon: 'text',
       text: {
-        dark: '#FFFFFF',
-        light: '#333333',
+        dark: '#DDDDDD',
```
ShimiSun, line 46: This change would break the accessibility contrast standard for primary Button.
https://webaim.org/resources/contrastchecker/?fcolor=DDDDDD&bcolor=01A982

PR #46, opened 2020-04-15
Latest text and formfield colors plus drop and select styling

halocline:
The following are updates reconciling text / formfield colors, as well as some drop and select specific styling. Links are to Figma files for reference.

- [text.dark](https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/hpe-color?node-id=0%3A1)
- [disabled-text](https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/hpe-color?node-id=0%3A1)
- [Drop Shadows](https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/hpe-color?node-id=405%3A25)
- [input text color](https://www.figma.com/file/fZJdRcNrvZ1JjdPRYHpFFU/hpe-select?node-id=103%3A140)
-[select options text size](https://www.figma.com/file/fZJdRcNrvZ1JjdPRYHpFFU/hpe-select?node-id=103%3A140)

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -241,17 +241,13 @@ export const hpe = deepFreeze({
     border: {
       radius: '4px',
     },
+    color: 'text-strong',
     padding: {
       vertical: '4px',
       horizontal: '10px',
     },
     extend: css`
       ${props => !props.plain && 'font-weight: bold;'}
-      ${({ primary, theme }) =>
-        primary &&
-        `color: ${
-          theme.global.colors['text-strong'][theme.dark ? 'dark' : 'light']
-        };`}
     `,
   },
   calendar: {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1,5 +1,6 @@
 // (C) Copyright 2020-2021 Hewlett Packard Enterprise Development LP
 import { css } from 'styled-components';
+import { FormDown, FormUp } from 'grommet-icons';
 
 const isObject = item =>
   item && typeof item === 'object' && !Array.isArray(item);
@@ -503,6 +504,8 @@ export const hpe = deepFreeze({
   select: {
     icons: {
       color: 'text',
+      down: FormDown,
+      up: FormUp,
     },
     options: {
       text: {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -169,18 +169,14 @@ export const hpe = deepFreeze({
       // Naming in Figma file is strong/default/weak vs. Grommet t-shirt sizing.
       // As defined here, default is currently mapping to medium.
       light: {
-        xsmall: '0px 1px 2px #0000003D;',
-        small: '0px 2px 4px #0000003D;',
-        medium: '0px 4px 8px #0000003D;',
-        large: '0px 6px 12px #0000003D;',
-        xlarge: '0px 12px 24px #0000003D;',
+        small: '0px 2px 4px #0000001F;',
+        medium: '0px 6px 12px #0000003D;',
+        large: '0px 12px 24px #0000001F;',
       },
       dark: {
-        xsmall: '0px 1px 2px #00000099;',
-        small: '0px 2px 4px #00000099;',
-        medium: '0px 4px 8px #00000099;',
-        large: '0px 6px 12px #00000099;',
-        xlarge: '0px 12px 24px #00000099;',
+        small: '0px 2px 4px #0000003D;',
+        medium: '0px 6px 12px #0000005C;',
+        large: '0px 12px 24px #0000007A;',
       },
     },
     hover: {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -43,7 +43,7 @@ export const hpe = deepFreeze({
       },
       icon: 'text',
       text: {
-        dark: '#FFFFFF',
+        dark: '#DDDDDD',
         light: '#444444',
       },
       'text-strong': {
@@ -65,8 +65,8 @@ export const hpe = deepFreeze({
       },
       'active-text': 'text',
       'disabled-text': {
-        dark: '#999999',
-        light: '#777777',
+        dark: '#777777',
+        light: '#999999',
       },
       'selected-background': 'brand',
       'selected-text': '#FFFFFF',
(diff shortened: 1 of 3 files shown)


src/js/index.js
```diff
@@ -492,10 +504,12 @@ export const hpe = deepFreeze({
   select: {
     icons: {
       color: 'text',
+      down: FormDown,
+      up: FormUp,
     },
     options: {
       text: {
-        size: 'small',
+        size: 'medium',
```
taysea, line 512: If we leave this undefined, will it use the default of medium?

src/js/index.js
```diff
@@ -157,11 +158,25 @@ export const hpe = deepFreeze({
       border: {
         radius: '4px',
       },
+      extend: ({ theme }) => `
+        margin-top: ${theme.global.edgeSize.xxsmall};
+      `,
       shadowSize: 'medium',
     },
     elevation: {
+      light: {
```
taysea, line 167: How were these different box shadow sizes calculated?

src/js/index.js
```diff
@@ -43,7 +44,7 @@ export const hpe = deepFreeze({
       },
       icon: 'text',
       text: {
-        dark: '#FFFFFF',
+        dark: '#DDDDDD',
```
taysea, line 47: I had held off on incorporating this change because buttons rely on `text` color. If we change this, then primary buttons go back to having the grey-ish text instead of pure white. Any thoughts on how to handle that case?

src/js/index.js
halocline: That would be my expectation, but didn't explicitly check.

src/js/index.js
halocline: There needs to be a namespace conversation. Figma styles reference 'strong' 'weak' etc, whereas Grommet uses t-shirt sizing. The "calculations" were just done mapping the shadow-default in [Figma](https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/HPE-Color-Styles?node-id=405%3A25) to 'medium' here, 'weak' to 'small', 'strong' to 'xlarge', then using the logical increments from there. The y-offsets and blur radii are very similar to Grommet (which has inconsistencies between light and dark). The Figma file is consistent across light/dark and Vicky and Greg confirmed.

src/js/index.js
halocline: Seems like something on the lines of this would be appropriate.
```
button: {
    color: {
      light: 'text-strong',
      dark: '#FFFFFF',
    },
  },
```
Doesn't feel right to have the published design colors and this value out of alignment, plus affects all text throughout...

src/js/index.js
taysea: Gotcha, this makes sense. Just hadn't seen them in a file and was curious. Agree that we should have a namespace discussion to align designs with grommet.

src/js/index.js
taysea: Gotcha -- I think we only want #FFFFFF if it's a primary button, but we could probably check that here. Agree we should align with published design colors but don't want a regression on the primary button color.

src/js/index.js
taysea: Cool, just wondering if we don't want to add unnecessary values to the theme if it's covered by default vs being exhaustive

src/js/index.js
ShimiSun: In the meanwhile, it will be helpful to add a comment here, so it won't block the progress of this PR.

src/js/index.js
ShimiSun: I agree it's unnecessary to state the object unless it overrides the base.js styles.

src/js/index.js
```diff
@@ -492,10 +504,12 @@ export const hpe = deepFreeze({
   select: {
     icons: {
       color: 'text',
+      down: FormDown,
+      up: FormUp,
```
ShimiSun, line 508: woot woot @taysea

src/js/index.js
halocline: removed.

src/js/index.js
halocline: comments added.

src/js/index.js
halocline: addressed by maintaining primary button text color via button.extend

src/js/index.js
halocline: woot woot @taysea, indeed

src/js/index.js
```diff
@@ -243,6 +247,11 @@ export const hpe = deepFreeze({
     },
     extend: css`
       ${props => !props.plain && 'font-weight: bold;'}
+      ${({ primary, theme }) =>
+        primary &&
+        `color: ${
+          theme.global.colors['text-strong'][theme.dark ? 'dark' : 'light']
+        };`}
```
ShimiSun, line 254: It would make more sense to use `button.primary.extend` https://v2.grommet.io/button#button.primary.extend instead of the general  `button.extend`

src/js/index.js
taysea: Oh you're right! Good catch on latest designs.

src/js/index.js
halocline: FYI, realized that the designs call for 'text-strong' across all buttons, not just primary, so making that adjustment.

src/js/index.js
```diff
@@ -157,11 +158,29 @@ export const hpe = deepFreeze({
...
     elevation: {
+      // Elevation values were derived from this Figma file.
+      // https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/HPE-Color-Styles?node-id=405%3A25
+      // Naming in Figma file is strong/default/weak vs. Grommet t-shirt sizing.
+      // As defined here, default is currently mapping to medium.
+      light: {
+        xsmall: '0px 1px 2px #0000003D;',
+        small: '0px 2px 4px #0000003D;',
+        medium: '0px 4px 8px #0000003D;',
+        large: '0px 6px 12px #0000003D;',
```
taysea, line 175: One more update actually, latest designs are showing 3 levels of elevation.
Should be:
small: '0px 2px 4px #0000003D;'
medium: '0px 6px 12px #0000003D;'
large: '0px 12px 24px #0000003D;

PR #47, opened 2020-04-17
Align Button styles to Figma file

taysea:
This aligns the disabled and active styling with what is shown here: https://www.figma.com/file/Oi5UEEQ33VCAyOKQcZ8Nr8/HPE-Button-Component

I did have one comment on the designs for @L0ZZI as to how active buttons would be visible if the background of the page matches the active-background color since that value is no longer using any opacity to create contrast. Left that as a comment on the Figma designs.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -62,14 +62,22 @@ export const hpe = deepFreeze({
         light: '#CCCCCC',
         dark: '#444444',
       },
+      'disabled-border': {
+        dark: '#777777',
+        light: '#999999',
+      },
       control: 'brand',
       'active-background': {
-        dark: '#FFFFFF1F',
-        light: '#CCCCCC99',
+        dark: 'background-front',
+        light: 'background-back',
       },
       'active-text': 'text-strong',
       'selected-background': 'brand',
       'selected-text': 'text-strong',
+      'disabled-text': {
+        dark: '#777777',
+        light: '#999999',
+      },
       'status-critical': '#FF4040',
       'status-warning': '#FFAA15',
       'status-ok': '#00C781',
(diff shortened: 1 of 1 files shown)


ericsoderberghp:
Should this be into the NEXT branch?

src/js/index.js
```diff
@@ -225,6 +233,25 @@ export const hpe = deepFreeze({
     extend: css`
       ${props => !props.plain && 'font-weight: bold;'}
     `,
+    disabled: {
+      border: {
+        color: 'disabled-border',
+      },
+      color: 'disabled-text',
+      opacity: 1.0,
+      extend: ({ primary, theme }) =>
```
ericsoderberghp, line 242: Any reason this can't be inside the `primary` part of the theme, rather than having to test for that here?

taysea: Since primary styling typically uses color as the fill for the button, I had set up primary disabled to use the disabled color as the fill for primary buttons and label color otherwise. 

Since designs call for uniform fill/label color regardless of button type, I did it this way. However, if we don't want `color` to control background fill for primary buttons we can discuss and update how grommet does the styling.

ericsoderberghp: In general, I'm trying have the theme not need to inspect props. To me, that feels like the theme structure is missing something. Here's the general theme structure approach:
```
button.primary.background
button.disabled.primary.background
```
So, we start with non-state keys and then we add top level keys for each state and replicate the non-state keys as needed under it.

taysea: Great, I think this is a more consistent convention. I will look into updating this on the Grommet side of things.

taysea:
Closing since theme structure has been updated. Will be opening a new PR shortly to apply new styles

PR #48, opened 2020-04-17
Allow input to respect font line-height

taysea:
Dependent on this grommet PR: https://github.com/grommet/grommet/pull/3974


ShimiSun:
I think this PR should be filed in the NEXT branch. otherwise LGTM.

PR #49, opened 2020-04-21
align checkbox theme with designs

britt6612:
## Work Completed 

- Aligned checkbox colors to match designs
- would depend on grommet [change](https://github.com/grommet/grommet/pull/3965) 

https://www.figma.com/file/7Mm1xDBTOtPHqggEVpaD2N/HPE-Checkbox-Component?node-id=1390%3A458

following would be added once grommet is merged in 

```
  checkbox: {
    box: {
      extend: ({ theme }) => `
        background-color: ${
          theme.global.colors.background[theme.dark ? 'dark' : 'light']
        };
        border-radius: 4px;
      `,
    },
}

```

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -272,8 +272,7 @@ export const hpe = deepFreeze({
         `,
     },
     icon: {
-      size: '18px',
-      extend: 'stroke: white;',
+      extend: 'stroke-width: 2px;',
     },
     gap: 'small',
     toggle: {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -248,13 +248,37 @@ export const hpe = deepFreeze({
     },
   },
   checkBox: {
-    color: 'selected-text',
-    gap: 'small',
+    border: {
+      color: 'border',
+      width: '1px',
+    },
     check: {
       radius: '2px',
+      extend: ({ theme, checked, indeterminate }) => `
+      ${(checked || indeterminate) && 'border: none;'}
+        ${(checked || indeterminate) &&
+          `background-color: ${theme.global.colors['green!']};`}
+        `,
     },
-    border: {
-      width: '1px',
+    icon: {
+      size: '18px',
+      extend: 'stroke: white;',
+    },
+    gap: 'small',
+    toggle: {
+      background: 'background',
+      color: 'background',
+      knob: {
+        extend: ({ theme }) => `
+           box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.12);
+           border: 1px solid ${
+             theme.global.colors.border[theme.dark ? 'dark' : 'light']
+           }
+        `,
+      },
+      extend: ({ checked, theme }) => `
+      ${checked && `background-color: ${theme.global.colors['green!']};`}
+    `,
     },
   },
   formField: {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -256,6 +256,7 @@ export const hpe = deepFreeze({
         color: 'background-contrast',
       },
     },
+    color: 'background',
     border: {
       color: 'border',
       width: '1px',
@@ -265,14 +266,15 @@ export const hpe = deepFreeze({
       extend: ({ theme, checked, indeterminate }) => `
       background-color: ${
         checked || indeterminate
-          ? theme.global.colors['green!']
+          ? theme.global.colors.green[theme.dark ? 'dark' : 'light']
           : theme.global.colors.background[theme.dark ? 'dark' : 'light']
       };
       ${(checked || indeterminate) && 'border: none;'}
         `,
     },
     icon: {
-      extend: 'stroke-width: 2px;',
+      extend: `stroke-width: 2px;
+      stroke: white;`,
     },
     gap: 'small',
     toggle: {
@@ -281,15 +283,18 @@ export const hpe = deepFreeze({
       knob: {
         extend: ({ theme }) => `
            box-shadow: ${
-              theme.global.elevation[theme.dark ? 'dark' : 'light'].small
-            };
+             theme.global.elevation[theme.dark ? 'dark' : 'light'].small
+           };
            border: 1px solid ${
              theme.global.colors.border[theme.dark ? 'dark' : 'light']
            }
         `,
       },
       extend: ({ checked, theme }) => `
-        ${checked && `background-color: ${theme.global.colors['green!']};`}
+        ${checked &&
+          `background-color: ${
+            theme.global.colors.green[theme.dark ? 'dark' : 'light']
+          };`}
       `,
     },
     extend: ({ theme }) => `
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -248,6 +248,14 @@ export const hpe = deepFreeze({
     },
   },
   checkBox: {
+    hover: {
+      border: {
+        color: undefined,
+      },
+      background: {
+        color: 'background-contrast',
+      },
+    },
     border: {
       color: 'border',
       width: '1px',
@@ -255,9 +263,12 @@ export const hpe = deepFreeze({
     check: {
       radius: '2px',
       extend: ({ theme, checked, indeterminate }) => `
+      background-color: ${
+        checked || indeterminate
+          ? theme.global.colors['green!']
+          : theme.global.colors.background[theme.dark ? 'dark' : 'light']
+      };
       ${(checked || indeterminate) && 'border: none;'}
-        ${(checked || indeterminate) &&
-          `background-color: ${theme.global.colors['green!']};`}
         `,
     },
     icon: {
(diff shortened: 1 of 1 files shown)


britt6612:
I was able to use the extend check to account for the `background` of the checkbox when checked vs when not checked. When it is not checked it will take the `background` color instead of being transparent . 

I also added in the extend a `width: 100%` This is so the checkbox container will fit horizontal  within the parent box/formfield.

I added in `pad 12px` since this is now going to be taken out of `formfield.content.pad` 
In Grommet formfield adds a 12px pad for `checkBox` `radiobuttongroup`  `rangeinput` 
since all 3 of these components will have their hover state within `formfield` we that extra padding from `formfield` out so the hover `background-contrast` can fill the containing `formfield` 

This now is only depended on adding in the background hover color for [grommet](https://github.com/grommet/grommet/pull/3995)

src/js/index.js
```diff
@@ -248,16 +248,58 @@ export const hpe = deepFreeze({
...
+      size: '18px',
+      extend: 'stroke: white;',
+    },
+    gap: 'small',
+    toggle: {
+      background: 'background',
+      color: 'background',
+      knob: {
+        extend: ({ theme }) => `
+           box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.12);
```
halocline, line 284: Trying to tie styling back to globals where appropriate:

```suggestion
           box-shadow: ${
              theme.global.elevation[theme.dark ? 'dark' : 'light'].small
            };
```

src/js/index.js
```diff
@@ -248,16 +248,58 @@ export const hpe = deepFreeze({
...
+           border: 1px solid ${
+             theme.global.colors.border[theme.dark ? 'dark' : 'light']
+           }
+        `,
+      },
+      extend: ({ checked, theme }) => `
+      ${checked && `background-color: ${theme.global.colors['green!']};`}
+    `,
```
halocline, line 292:
```suggestion
      extend: ({ checked, theme }) => `
        ${checked && `background-color: ${theme.global.colors['green!']};`}
      `,
```

src/js/index.js
```diff
@@ -248,16 +248,58 @@ export const hpe = deepFreeze({
...
+    },
+    extend: `
+    width: 100%;
+    padding: 12px;
+`,
```
halocline, line 297:
```suggestion
    extend: `
      width: 100%;
      padding: 12px;
    `,
```

src/js/index.js
```diff
@@ -248,16 +248,60 @@ export const hpe = deepFreeze({
...
+      border: {
+        color: undefined,
+      },
+      background: {
+        color: 'background-contrast',
+      },
     },
     border: {
+      color: 'border',
       width: '1px',
```
halocline, line 261:
```suggestion
      width: 'xsmall',
```

src/js/index.js
```diff
@@ -248,16 +248,60 @@ export const hpe = deepFreeze({
...
     },
+    check: {
+      radius: '2px',
```
halocline, line 264: I'd like to be using 'xxsmall' here, but we need to make adjustments to our global.edgeSizes

src/js/index.js
```diff
@@ -248,16 +248,60 @@ export const hpe = deepFreeze({
...
+        checked || indeterminate
+          ? theme.global.colors['green!']
+          : theme.global.colors.background[theme.dark ? 'dark' : 'light']
+      };
+      ${(checked || indeterminate) && 'border: none;'}
+        `,
+    },
+    icon: {
+      size: '18px',
+      extend: 'stroke: white;',
```
halocline, line 276: Is this necessary? We may be able to remove.

britt6612: The white yes we wont need the size let me double check without this it was `24px` which seemed bigger then designs but I will check and make sure first!

src/js/index.js
britt6612: I agree I didn't like putting values but there was no global sizes

src/js/index.js
```diff
@@ -248,16 +248,60 @@ export const hpe = deepFreeze({
...
+      background-color: ${
+        checked || indeterminate
+          ? theme.global.colors['green!']
+          : theme.global.colors.background[theme.dark ? 'dark' : 'light']
+      };
+      ${(checked || indeterminate) && 'border: none;'}
+        `,
+    },
+    icon: {
+      size: '18px',
```
halocline, line 275: What are we trying to achieve here? It seems like the icon should scale to its parent.

britt6612: I was trying to make the icon smaller but turns out the size was fine it was the stroke width that was wrong it was `4px` by default and in designs it is `2px`

src/js/index.js
halocline: Can you make an issue for "Create Global Edge Sizes for HPE Theme" and link it to this?

src/js/index.js
halocline: 👍

src/js/index.js
britt6612: yes on it! 👍

britt6612:
This would still be waiting on the hover background PR in Grommet

src/js/index.js
```diff
@@ -248,16 +248,59 @@ export const hpe = deepFreeze({
...
+           }
+        `,
+      },
+      extend: ({ checked, theme }) => `
+        ${checked && `background-color: ${theme.global.colors['green!']};`}
+      `,
+    },
+    extend: ({ theme }) => `
+      width: 100%;
+      padding: ${theme.global.edgeSize.small};
```
taysea, line 297: Confirming that this adds back in the pad that is removed in 301? and by doing this, the hover background fills the entire perceived formfield background @halocline

halocline: Correct. Removing content.pad from FormField which is only applied to CheckBox, RadioButtonGroup, RangeInput. Instead, adding pad to CheckBox for background fills.

Relevant Grommet:
https://github.com/grommet/grommet/blob/2eb8d09df21c2e6bfc86bd069e5bfaa1e9193ab9/src/js/components/FormField/FormField.js#L21
https://github.com/grommet/grommet/blob/2eb8d09df21c2e6bfc86bd069e5bfaa1e9193ab9/src/js/components/FormField/FormField.js#L174

britt6612: This was the direction we decided on. 👍

src/js/index.js
ShimiSun: Along those lines
```suggestion
      radius: '2px', // should use 'xxsmall' of global.edgeSizes
```

src/js/index.js
```diff
@@ -248,16 +248,59 @@ export const hpe = deepFreeze({
...
+    `,
   },
   formField: {
+    content: {
+      pad: undefined,
+    },
```
ShimiSun, line 303: @halocline, are we ok with this change affecting all forms and not only the checkbox? is this how we will recommend consuming the component without any pad? or is it applied elsewhere and I'm missing something.

src/js/index.js
```diff
@@ -248,16 +248,59 @@ export const hpe = deepFreeze({
...
     },
+    check: {
+      radius: '2px',
+      extend: ({ theme, checked, indeterminate }) => `
+      background-color: ${
+        checked || indeterminate
+          ? theme.global.colors['green!']
+          : theme.global.colors.background[theme.dark ? 'dark' : 'light']
+      };
+      ${(checked || indeterminate) && 'border: none;'}
```
ShimiSun, line 271: Can you confirm this functionality was tested on the design-system and it works as expected for all cases?

britt6612: I did I can push it up on my checkbox branch if anyone wanted to check I just kept it locally on my end.

src/js/index.js
halocline: My understanding is that this `content.pad` affects CheckBox, RadioButtonGroup, and RangeInput. Each of these will need the small pad re-applied which will account for the spacing and the background color for the active/hover states. For CheckBox, this has been applied on Line 297. RadioButtonGroup and RangeInput will need a similar approach.

ericsoderberghp: I'm not completely up to date on the details here, but I think CheckBoxGroup, RadioButtonGroup, and RangeInput should all have the right pad built into them such that they align with the other input elements.

halocline: They do. However, the designs call for these inputs to occupy more space, most apparent in the active/hover states. The approach Brittany is proposing here is to transfer the correct padding from FormField and instead apply it on the input.
Screen Shot 2020-04-24 at 9 30 03 AM

britt6612: Since the padding is coming from `formfield` It looked like the picture below where the hover color was not covering the whole component as in the designs.  I purposed that we took that padding out and added it back into the checkbox component itself, so that way the hover color was able to fill the box around it. 
We can also discuss other ways around this. This was just one way I had thought of. 

Screen Shot 2020-04-22 at 3 38 24 PM

ericsoderberghp: Totally agreed! The padding should be on CheckBoxGroup and friends, not FormField.

britt6612: yay 😃sounds good!

ShimiSun:
I confirmed this PR isn't pending on grommet's PRs and that it was tested with the design-system. Thank you!

PR #50, opened 2020-04-21
Update colors based on latest swatches

taysea:
### In this PR
This updates color values/names as well as tab styling (since some is reliant on these values). I am updating everything that has a 1:1 mapping or could easily added but outstanding issues are noted at the bottom of this PR.

The values are coming from here: https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/HPE-Color-Styles?node-id=590%3A981

### Items that need to be addressed separately:
- [ ] How will these new validation colors play out in a form? I believe it will require some grommet changes, but I'm curious if the styles will be opt-in depending on the form or if they'll automatically be applied: 
Screen Shot 2020-04-21 at 3 56 20 PM

- [ ] formField.disabled.background has not been updated. Currently, global.opacity.medium is being applied to disabled formFields, so applying a color of `#0000000f` gets over-ridden by rgba(0,0,0,0.4) in the DOM. I haven't filled in this value because it will make disabled fields too dark. Need to look into fix more.
- [ ] Control colors of primary/secondary need to be discussed a little more -- are these just for Button or will this apply to other control components like menu, etc. where a control color is sometimes used?

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -548,8 +548,8 @@ export const hpe = deepFreeze({
     pad: 'small',
     margin: 'none',
     extend: ({ theme }) => css`
-      border-top-left-radius: ${theme.global.control.border.radius};
-      border-top-right-radius: ${theme.global.control.border.radius};
+      border-top-left-radius: ${theme.global.control.border.radius}; // should use radius property of border
+      border-top-right-radius: ${theme.global.control.border.radius}; // should use radius property of border
       font-weight: bold;
     `,
   },
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -537,12 +537,12 @@ export const hpe = deepFreeze({
     },
     border: {
       side: 'bottom',
-      color: 'background-back',
+      color: 'border',
       active: {
-        color: 'border',
+        color: 'border-strong',
       },
       hover: {
-        color: 'background-back',
+        color: 'border',
       },
     },
     pad: 'small',
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -26,24 +26,24 @@ export const hpe = deepFreeze({
       'neutral-4': undefined,
       brand: 'green!',
       background: {
-        dark: '#1A1F2B',
+        dark: '#263040',
         light: '#FFFFFF',
       },
       'background-back': {
-        dark: '#1A1F2B',
+        dark: '#263040',
         light: '#EFEFEF',
       },
       'background-front': {
-        dark: '#354050',
+        dark: '#404B5C',
         light: '#FFFFFF',
       },
       'background-contrast': {
-        dark: '#FFFFFF1F',
-        light: '#CCCCCC99',
+        dark: '#FFFFFF14',
+        light: '#0000000A',
       },
       icon: 'text',
       text: {
-        dark: '#FFFFFF',
+        dark: '#C0CADC',
         light: '#444444',
       },
       'text-strong': {
@@ -51,12 +51,20 @@ export const hpe = deepFreeze({
         light: '#000000',
       },
       'text-weak': {
-        dark: '#AAAAAA',
-        light: '#666666',
+        dark: '#606B7D',
+        light: '#BBBBBB',
       },
       border: {
-        dark: '#DDDDDD',
+        dark: '#7887A1',
+        light: '#999999',
+      },
+      'border-strong': {
+        dark: '#AFBCD2',
         light: '#666666',
+      },
+      'border-weak': {
+        dark: '#606B7D',
+        light: '#BBBBBB',
       },
       control: 'brand',
       'active-background': {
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -70,45 +78,51 @@ export const hpe = deepFreeze({
...
-        light: '#FFB024',
+        dark: '#9B6310',
+        light: '#FFBC44',
       },
       'orange!': '#FF8300',
       yellow: {
-        dark: '#D78F00',
-        light: '#FFEB59',
+        dark: '#8D741C',
+        light: '#FFD63E',
```
halocline, line 125: Looks like this has shifted: https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/HPE-Color-Styles?node-id=590%3A1231
```suggestion
        light: '#FFEB59',
```

src/js/index.js
```diff
@@ -286,7 +300,7 @@ export const hpe = deepFreeze({
     },
     error: {
       background: {
-        color: { light: '#FF404033', dark: '#FF40404D' },
+        color: { light: '#FC6161', dark: '#C54E4B' },
```
halocline, line 303: These look suspect. They are the colors of 'red' and 'red!', which are correct-ish. The visual color has these colors plus a 24% and 36% opacity. Ideally we would reference by name rather than use HEX here. Potentially could have a function which receives (colorname, opacity) and returns a hexvalue with opacity.

```suggestion
        color: { light: '#FC61613D', dark: '#C54E4B5C' },
```

src/js/index.js
```diff
@@ -513,16 +527,31 @@ export const hpe = deepFreeze({
...
-      horizontal: 'small',
+    border: {
+      side: 'bottom',
+      color: 'background-back',
+      active: {
+        color: 'border',
+      },
+      hover: {
+        color: 'background-back',
+      },
```
halocline, line 546: Seems like these may have changed. In https://www.figma.com/file/Kp4dWyhUTnKIJ1Cg5CoL9o/HPE-Tabs-Component?node-id=208%3A0 I am seeing the following:

```suggestion
      color: 'border',
      active: {
        color: 'border-strong',
      },
      hover: {
        color: 'border',
      },
```

src/js/index.js
halocline: This may be related to your commentary in the PR description, but wasn't exactly sure.

src/js/index.js
```diff
@@ -513,16 +527,31 @@ export const hpe = deepFreeze({
...
+      hover: {
+        color: 'border',
+      },
     },
+    pad: 'small',
+    margin: 'none',
+    extend: ({ theme }) => css`
+      border-top-left-radius: ${theme.global.control.border.radius};
+      border-top-right-radius: ${theme.global.control.border.radius};
```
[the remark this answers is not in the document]
taysea: Sounds good -- is that something we need to implement in grommet prior to merging this PR? or something that can be added in after and then updated?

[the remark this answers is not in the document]
ShimiSun: we can add it later. for now:
```suggestion
      border-top-left-radius: ${theme.global.control.border.radius}; // should use radius property of border
      border-top-right-radius: ${theme.global.control.border.radius}; // should use radius property of border
```

src/js/index.js
```diff
@@ -27,38 +27,46 @@ export const hpe = deepFreeze({
...
+        light: '#999999',
+      },
+      'border-strong': {
+        dark: '#AFBCD2',
         light: '#666666',
       },
+      'border-weak': {
+        dark: '#606B7D',
+        light: '#BBBBBB',
+      },
```
ShimiSun, line 69: I commented it earlier and didn't see a direct reply so reposting
>Just to confirm, were the new colors tested on the application side prior to the PR submission? Just to be sure we are not introducing new contrast issues.

taysea: Bringing over a discussion from Slack:
Using text color of `text` on the dark colors doesn’t meet contrast requirements, but changing the text color to `text-strong` passes everything. Feels like the site will need some guidance of when to use text vs text-strong to avoid the first error

`text` on dark color palette:
Screen Shot 2020-04-23 at 10 04 59 AM

`text-strong` on dark color palette:
Screen Shot 2020-04-23 at 10 13 01 AM

taysea: As long as we provide appropriate guidance of how to use text colors, then we are fine from a color contrast perspective.

src/js/index.js
```diff
@@ -513,16 +527,31 @@ export const hpe = deepFreeze({
...
+    hover: {
+      background: 'background-contrast',
+      color: 'text',
     },
-    pad: {
-      horizontal: 'small',
+    border: {
+      side: 'bottom',
+      color: 'border',
+      active: {
```
ericsoderberghp, line 541: On the grommet side, this should be `active.border` instead of `border.active` to align with the theme structure direction. Not an issue for this PR per-se.

src/js/index.js
```diff
@@ -513,16 +527,31 @@ export const hpe = deepFreeze({
...
+        color: 'border-strong',
+      },
+      hover: {
+        color: 'border',
+      },
     },
+    pad: 'small',
+    margin: 'none',
+    extend: ({ theme }) => css`
```
ericsoderberghp, line 550: Would be nice to not need `extend` for HPE theme. Not part of this PR per-se, but something we should look into. Can we add an issue to evaluate the use of `extend` in the HPE NEXT theme, to see if anything could/should be easier.

PR #61, opened 2020-05-01
add in margin & hover to accordion

britt6612:
Adding to margin for accordion and undefined for hover color change.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -207,7 +207,7 @@ export const hpe = deepFreeze({
   },
   accordion: {
     heading: {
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -206,6 +206,12 @@ export const hpe = deepFreeze({
     },
   },
   accordion: {
+    heading: {
+      margin: { vertical: '24px', left: '6px', right: '6px', },
```
halocline, line 210: Can these be done with t-shirt sizes?

britt6612: good point and ill do horizontal 6px not both left and right

PR #69, opened 2020-05-12
update margins based on Slack thread 5-12-2020

halocline:
[Figma](https://www.figma.com/file/fZJdRcNrvZ1JjdPRYHpFFU/HPE-Select-Component?node-id=103%3A140)

#### What does this PR do?

Updates FormField label and help text margins to complete alignment with Figma - [Issue](https://github.com/hpe-design/design-system/issues/818).

Figma design calls for no spacing between label and help text. Previously, bottom margin was set in the theme at the label, help, error, info levels. This created unwanted spacing when label and help were together. This removes the margin from those elements and transfers the desired vertical margin to the FormFieldContentBox instead.

Yes.

Applied to aries.js on the Design System site. Screenshots below.

This approach was dependent on a change in Grommet FormField to allow for margin to be specified via theme [PR](https://github.com/grommet/grommet/pull/4148).

https://github.com/hpe-design/design-system/issues/818
[Slack thread](https://hpe.slack.com/archives/GPPLUK6LR/p1589304674002700)

**Before** -- Note the unwanted spacing between the label and the help text.
Screen Shot 2020-08-11 at 9 00 42 AM

**After** -- Unwanted spacing eliminated. Spacing between formField content box and label/help/error/info text maintained.
Screen Shot 2020-08-11 at 9 16 56 AM

This would be a breaking change when we merge NEXT into Master, it modifies the spacing between FormField label texts.

Implemented FormField label and input spacing to align with Figma designs.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -466,10 +466,19 @@ export const hpe = deepFreeze({
       },
       size: 'xsmall',
       color: 'text',
-      margin: {
-        start: 'none',
-      },
+      margin: 'none',
     },
+    extend: `
+      // This selector will work when label or help
+      // are passed strings. However, it will be too
+      // broad in situations where a node is passed.
+      // The proper solution should be supporting a
+      // formField.input.margin or formField.input in 
+      // the theme.
+      > div:first-of-type {
+        margin: 4px 0px;
+      }
+    `,
     focus: {
       border: {
         color: 'border-strong',
@@ -478,17 +487,12 @@ export const hpe = deepFreeze({
     help: {
       size: 'xsmall',
       color: 'text',
-      margin: {
-        start: 'none',
-        bottom: '4px',
-      },
+      margin: 'none',
     },
     info: {
       size: 'xsmall',
       color: 'text',
-      margin: {
-        start: 'none',
-      },
+      margin: 'none',
     },
     label: {
       size: 'xsmall',
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -486,6 +486,7 @@ export const hpe = deepFreeze({
   },
   formField: {
     content: {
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -493,9 +493,7 @@ export const hpe = deepFreeze({
     label: {
       size: 'xsmall',
       color: 'text',
-      margin: {
-        horizontal: 'none',
-      },
+      margin: 'none',
```
taysea, line 496: Is the spacing between label and formfield still correct in cases when there is no help text?

halocline: Good call.

src/js/index.js
```diff
@@ -466,10 +466,19 @@ export const hpe = deepFreeze({
...
     },
+    extend: `
+      // This selector will work when label or help
+      // are passed strings. However, it will be too
+      // broad in situations where a node is passed.
+      // The proper solution should be supporting a
+      // formField.input.margin or formField.input in 
+      // the theme.
+      > div:first-of-type {
+        margin: 4px 0px;
```
taysea, line 479: It seems like what we really want is to allow formField to have a margin-top defined and then help and label would both have no margin. That way, regardless of if just a label or label + help were provided, the correct spacing between text and formField would apply. This would be a grommet change though because I don't think that's currently supported.

ericsoderberghp: This does look like the kind of extend behavior that we'd like to not need.

halocline: Correct and agree.

Was primarily using this as documentation of my investigation so that it could be referenced for Grommet enhancement.

ericsoderberghp: yep, well done 👍

PR #77, opened 2020-06-09
Update tabs so border extends

taysea:
Updating tabs styling based on new grommet theming capabilities which allows us to close https://github.com/hpe-design/design-system/issues/789

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -743,7 +743,11 @@ export const hpe = deepFreeze({
       },
     },
     pad: 'small',
-    margin: 'none',
+    margin: {
+      // bring the overall tabs border behind invidual tab borders
+      vertical: '-2px',
+      horizontal: 'none',
+    },
     extend: ({ theme }) => css`
       border-top-left-radius: ${theme.global.control.border
         .radius}; // should use radius property of border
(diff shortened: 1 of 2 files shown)


src/js/index.js
```diff
@@ -743,7 +743,11 @@ export const hpe = deepFreeze({
       },
     },
     pad: 'small',
-    margin: 'none',
+    margin: {
+      // bring the overall tabs border behind invidual tab borders
+      vertical: '-2px',
```
ShimiSun, line 748: feels hacky

taysea: It is a little bit hacky, but the reason we have to do it this way is because the tabs container wraps around the outside of the individual tabs. So, in order the have this border extend inline with the individual tab borders, we have to use a negative margin on the individual tabs. @ericsoderberghp and I chatted about this approach and found it to be better than having to do theming in the extend.

This allows us to achieve this by default (where the tabs bottom border will span the width of its container):
Screen Shot 2020-06-09 at 1 00 23 PM

For use cases (if they arise), where the user does not want this extended border, they can use `alignControls=start` to achieve this, where the controls alignSelf to the start, but the tab content would still be able to span the full width:
Screen Shot 2020-06-09 at 1 03 50 PM

Eric noted that it would be fine for the default to be that the border extends the full width.

ShimiSun: ok

taysea: Unless you think there would be a better way to move that `Tabs` border to the same level as the other borders. I am open to discussing more grommety ways if you have something in mind!

PR #87, opened 2020-07-31
add error validation to color

britt6612:
#### What does this PR do?
This PR adds the color we are using for `formfield` to colors
aries-site
Needed a new color namespace in order to call easily 
https://github.com/hpe-design/design-system/issues/1009

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -118,7 +118,7 @@ export const hpe = deepFreeze({
         light: '#FFBC44',
       },
       'orange!': '#FF8300',
-      'validation-error': {
+      'validation-critical': {
         light: '#FC61613D',
         dark: '#C54E4B5C',
       },
@@ -510,7 +510,7 @@ export const hpe = deepFreeze({
     },
     error: {
       background: {
-        color: 'validation-error',
+        color: 'validation-critical',
       },
       size: 'xsmall',
       color: 'text',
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -118,6 +118,10 @@ export const hpe = deepFreeze({
         light: '#FFBC44',
       },
       'orange!': '#FF8300',
+      'validation-error': {
```
taysea, line 121: I'm noticing that this color name isn't aligned between the site and Figma. In figma, we show a few validation colors but they're all marked "validation". Let's get resolution from @gregfuruiye or another designer and update the site or Figma where necessary before merging so all of these match

gregfuruiye: Since we are calling most things red 'critical', I wonder if that's what we stick with for anything giving red status. There is another use case where the validation color is being used in notifications for critical alerts, so I'm thinking we just call it validation-critical:

status-critical
validation-critical

britt6612: I think that approach is good! We should just line up figma files with the name

src/js/index.js
```diff
@@ -462,10 +466,12 @@ export const hpe = deepFreeze({
         `,
       },
       extend: ({ checked, theme }) => `
-        ${checked &&
+        ${
```
taysea, line 469: Was this a linter fix or unwanted change?

PR #97, opened 2020-08-13
add size for dateinput icon

britt6612:
#### What does this PR do?
This PR adds a size for the calendar icon 
yes
visual
The icon used in grommet is to large for what is in the designs 
issue #1042 in design system
backward compatible

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -181,6 +181,11 @@ export const hpe = deepFreeze({
       background: 'active-background',
       color: 'active-text',
     },
+    dateInput: {
+      icon: {
+        size: 'small',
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -181,6 +181,11 @@ export const hpe = deepFreeze({
       background: 'active-background',
       color: 'active-text',
     },
+    dateInput: {
```
halocline, line 184: This looks like it is in global. Does it need to be move after checkBox and before formField?

britt6612: oh yes just moved!

PR #98, opened 2020-08-14
Added interim theme

ShimiSun:
#### What does this PR do?
This PR suggests an interim theme that will mitigate the breaking changes to the design-system theme upgrade.
Yes, done.
None, yet.
Follow up on our GLC migration.
backward compatible
The design-system is providing an interim theme that helps mitigate the breaking changes of the theme upgrade. 
More notes on where the interim theme could be found will be helpful as well.

### DIFF
--- a/src/js/interim.js
+++ b/src/js/interim.js
@@ -1,0 +1,31 @@
+/**
+ * The following theme has transitional properties to
+ * mitigate the upgrade to the design-system theme.
+ * Using the interim theme will help you avoid critical
+ * and immediate breaking changes.
+ *
+ * How does it work?
+ * Copy the content of the interimTheme file to your local theme file
+ * and update your official HPE theme to point to the new HPE
+ * design-system theme (via your package.json).
+ * Copying the content of this file will help you avoid immediate
+ * breaking changes behavior.
+ * However the main goal is for your theme to work independently without
+ * the interim file support.
+ *
+ */
+
+export const interimTheme = {
+  global: {
+    colors: {
+      'accent-1': '#6FFFB0',
+      'accent-2': '#FD6FFF',
+      'accent-3': '#81FCED',
+      'accent-4': '#FFCA58',
+      'neutral-1': '#00873D',
+      'neutral-2': '#3D138D',
+      'neutral-3': '#00739D',
+      'neutral-4': '#A2423D',
+    },
+  },
+};
--- a/src/js/interim.js
+++ b/src/js/interim.js
@@ -1,6 +1,7 @@
 /**
  * The following theme has transitional properties to
- * mitigate the upgrade to the design-system theme.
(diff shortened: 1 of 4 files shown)


ericsoderberghp:
how about naming it something like "v2migration", just to make it a bit clearer?

src/js/interim.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * The following theme has transitional properties to
+ * mitigate the upgrade to the design-system theme.
```
ericsoderberghp, line 3: I think we should be specific about migrating from v1 to v2 of the theme.

src/js/interim.js
```diff
@@ -0,0 +1,31 @@
...
+ * Using the interim theme will help you avoid critical
+ * and immediate breaking changes.
+ *
+ * How does it work?
+ * Copy the content of the interimTheme file to your local theme file
```
ericsoderberghp, line 8: Rather than "Copy", can we set it up so that:
`import { v2migration } from 'grommet-theme-hpe';`
works?

ShimiSun: ...And then you are getting into a potential 3 layer theming (hpe+v5Migration+product themes), and you need to document how to use the deepMerge in between the layers. 
I thought about it too at the beginning and decided to keep the 'Copy' instead.
I don't think this theme will be used much, so I prefer to keep the usage of it simple.

src/js/interim.js
ShimiSun: Done.

src/js/interim.js
ericsoderberghp: Why does the caller need to do `deepMerge`? I was trying to keep the caller's experience as simple as possible:
```
import { v2migration } from 'grommet-theme-hpe';
<Grommet theme={v2migration} />
```
which would allow us to do anything we needed to inside the `v2migration` theme.

ShimiSun: In your example, you are assuming there is no local theme for the specific product, which is not the case in most scenarios.

In addition, What about the hpe theme? 
```
import { v2migration, hpe } from 'grommet-theme-hpe';
```
we still want the user to import the hpe theme, so it will be easier to separate the themes and understand each theme's role while gradually eliminating styles from the `v2migration` theme.
Eliminating the styles gradually from the `v2migration` theme will be easier if you copy the styles to your local file and take them off one by one.

ericsoderberghp: I'm assuming any local theme is handled the same way it is handled today, perhaps merging with theme from grommet-theme-hpe. So, no change in that regard. And, I'm assuming that `v2migration` has the `hpe` theme included in it. So, you use `v2migration` until you are done and then switch to the latest `hpe`. But, I'm happy to defer to you on this, as you've been helping the Green Lake team out directly.

PR #101, opened 2020-08-18
add margin to drop

britt6612:
#### What does this PR do?
This PR adds margin for top and bottom of drop 
yes
design system to match figma files
https://github.com/hpe-design/design-system/issues/657
issue #657 on design system

Screen Shot 2020-08-18 at 12 11 41 PM
Screen Shot 2020-08-18 at 12 11 55 PM
Screen Shot 2020-08-18 at 12 12 22 PM

Screen Shot 2020-08-18 at 12 11 32 PM

backward compatible
This should be added that previously there was no margin between `textinput` and `drop` and now it is 6px

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -186,6 +186,9 @@ export const hpe = deepFreeze({
       border: {
         radius: '4px',
       },
+      extend: ({ theme }) =>
+        `margin-top: ${theme.global.edgeSize.xsmall}; 
+       margin-bottom: ${theme.global.edgeSize.xsmall};`,
       shadowSize: 'medium',
     },
     elevation: {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -186,9 +186,11 @@ export const hpe = deepFreeze({
       border: {
         radius: '4px',
       },
-      extend: ({ theme }) =>
-        `margin-top: ${theme.global.edgeSize.xsmall}; 
-       margin-bottom: ${theme.global.edgeSize.xsmall};`,
+      extend: ({ alignProp, theme }) =>
+        `margin-top: ${alignProp.top !== 'top' &&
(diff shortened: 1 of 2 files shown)


src/js/index.js
```diff
@@ -186,6 +186,9 @@ export const hpe = deepFreeze({
       border: {
         radius: '4px',
       },
+      extend: ({ theme }) =>
+        `margin-top: ${theme.global.edgeSize.xsmall};
```
taysea, line 190: For future reference, `theme.global.edgeSize.xsmall` = `6px` which aligns with what is shown in designs.

src/js/index.js
```diff
@@ -186,6 +186,9 @@ export const hpe = deepFreeze({
       border: {
         radius: '4px',
       },
+      extend: ({ theme }) =>
+        `margin-top: ${theme.global.edgeSize.xsmall}; 
+       margin-bottom: ${theme.global.edgeSize.xsmall};`,
```
halocline, line 191: Just a formatting suggestion.

```suggestion
      extend: ({ theme }) => `
        margin-top: ${theme.global.edgeSize.xsmall};
        margin-bottom: ${theme.global.edgeSize.xsmall};
      `,
```

britt6612:
> Realizing the behavior is a little weird for cases like `Menu` where the button overlays when the Menu opens but gets shifted down by the 6px. seems like in that case of menu we wouldn't want this margin to apply.

I. just noticed that as well that you pointed it out. It didnt stand out to me before

britt6612:
Let me check if there is a way I can exclude menu from having that margin

britt6612:
@taysea thought of a great idea with just ignoring menu.
In grommet there is default values for `align-top` and `align-bottom` 
since we know what those are we can check to see if those are set aka if its `menu` and not set the margin

PR #108, opened 2020-09-02
Merge NEXT to master branch

ShimiSun:
This PR is our preparation for releasing the official HPE Design-System theme.


ericsoderberghp:
I don't think we should merge this with `"version": "1.0.5",` still in package.json. Won't that cause a republish with this content?

ericsoderberghp (approved):

PR #104, opened 2020-08-26
Use radioButton.container

taysea:
#### What does this PR do?

Utilizes `radioButton.container.extend` for proper theming.

#### Should this PR be placed on the NEXT branch (design-system theme)?
Yes.

#### What testing has been done on this PR?
Test in design-system site.

#### Any background context you want to provide?
Grommet was applying `radioButton.extend` to two DOM elements but is not creating a separate theme object for `radioButton.container.extend`

#### What are the relevant issues?
https://github.com/grommet/grommet/pull/4361

#### Is this change backward compatible or could it be a breaking change for the official HPE theme?
Breaking change.

#### How should this PR be communicated in the release notes?
Already documented in [theme migration guide.](https://github.com/grommet/grommet-theme-hpe/wiki/Migration-guide-to-the-design-system-theme#radiobutton). However, this functionality is dependent on the merge of https://github.com/grommet/grommet/pull/4361 in grommet core.

PR #120, opened 2020-10-08
Use radioButton.container

taysea:
#### What does this PR do?
Move appropriate radiobutton.extend theming to radiobutton.container.extend.

tested locally on design system site.


ShimiSun:
Linking to #104 and grommet/grommet#4361

PR #129, opened 2020-10-28
Restraining the use of text-xweak

ShimiSun:
#### What does this PR do?
The HPE design-system is providing three options for text colors: `text`, `text-weak`, `text-xweak`.
Though, HPE users still see the `text-xweak` option from the base theme. This PR is restricting the use of this base theme color.

This PR results from a question we've received from GLC:
> I noticed that there is a text-xweak defined in the base grommet theme, but not in the hpe theme - and it is darker than the new text-weak is. Should the hpe theme also be specifying text-xweak so that it is relatively weaker than text-weak?

Might be an issue for HPE users that have used `text-xweak`, the value will be turned to 'undefined'.

Restricting the usage of `text-xweak` as this color isn't supported by the HPE design-system.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -30,6 +30,7 @@ export const hpe = deepFreeze({
       'neutral-4': undefined,
       'neutral-5': undefined,
       'status-error': undefined,
+      'text-xweak': undefined,
       brand: 'green!',
       background: {
         dark: '#263040',
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -30,7 +30,6 @@ export const hpe = deepFreeze({
       'neutral-4': undefined,
       'neutral-5': undefined,
       'status-error': undefined,
-      'text-xweak': undefined,
       brand: 'green!',
       background: {
         dark: '#263040',
@@ -61,6 +60,7 @@ export const hpe = deepFreeze({
         dark: '#606B7D',
         light: '#BBBBBB',
       },
+      'text-xweak': 'text-weak',
       border: {
         dark: '#7887A1',
         light: '#999999',


ericsoderberghp:
DataChart Legend uses `text-xweak`. Instead of removing it, can we give it a color value?

taysea:
> DataChart Legend uses `text-xweak`. Instead of removing it, can we give it a color value?

Just to make sure I'm understanding, are you suggesting we update DataChart to use a color value other than `text-xweak`? or suggesting we define a `text-xweak` specific for hpe theme?

ericsoderberghp:
> > DataChart Legend uses `text-xweak`. Instead of removing it, can we give it a color value?
> 
> Just to make sure I'm understanding, are you suggesting we update DataChart to use a color value other than `text-xweak`? or suggesting we define a `text-xweak` specific for hpe theme?

define a `text-xweak` specific for hpe theme

taysea:
> > > DataChart Legend uses `text-xweak`. Instead of removing it, can we give it a color value?
> > 
> > 
> > Just to make sure I'm understanding, are you suggesting we update DataChart to use a color value other than `text-xweak`? or suggesting we define a `text-xweak` specific for hpe theme?
> 
> define a `text-xweak` specific for hpe theme

Gotcha. I'd want to touch base with designers on that because I thought there was motivation to limit the number of tiers of text/border colors to avoid too many variations in color usage. Additionally, if we define a `text-xweak` it would need to be lighter than our `text-weak` but I believe as of now the grommet `text-xweak` that's being pulled is still darker than hpe `text-weak`, so would the usage of a lighter `text-xweak` as an hpe specific value still be accessible for the use case it's in for DataChart? Or would it make more sense for DataChart to use `text-weak` instead?

ericsoderberghp:
> Gotcha. I'd want to touch base with designers on that because I thought there was motivation to limit the number of tiers of text/border colors to avoid too many variations in color usage. Additionally, if we define a `text-xweak` it would need to be lighter than our `text-weak` but I believe as of now the grommet `text-xweak` that's being pulled is still darker than hpe `text-weak`, so would the usage of a lighter `text-xweak` as an hpe specific value still be accessible for the use case it's in for DataChart? Or would it make more sense for DataChart to use `text-weak` instead?

I would suggest that minimally, we would set `text-xweak` to the same value as `text-weak` in the HPE theme. This is somewhat similar to making effectively the same change in DataChart, but the DataChart change would be breaking. Changing just the HPE theme allows for themes to use the more nuanced grommet text levels.

taysea:
> > Gotcha. I'd want to touch base with designers on that because I thought there was motivation to limit the number of tiers of text/border colors to avoid too many variations in color usage. Additionally, if we define a `text-xweak` it would need to be lighter than our `text-weak` but I believe as of now the grommet `text-xweak` that's being pulled is still darker than hpe `text-weak`, so would the usage of a lighter `text-xweak` as an hpe specific value still be accessible for the use case it's in for DataChart? Or would it make more sense for DataChart to use `text-weak` instead?
> 
> I would suggest that minimally, we would set `text-xweak` to the same value as `text-weak` in the HPE theme. This is somewhat similar to making effectively the same change in DataChart, but the DataChart change would be breaking. Changing just the HPE theme allows for themes to use the more nuanced grommet text levels.

I think the suggestion of defining `text-xweak` as the same value as `text-weak` seems like the best approach for now as a way to put a sort of floor on the text values. What do you think @ShimiSun ?

ShimiSun:
Yes, I agree. Thanks!

taysea (approved):
LGTM.

ericsoderberghp (approved):

ShimiSun:
Thanks, I've updated the PR description & title to reflect the change better.

PR #133, opened 2020-11-02
Mark disabled-text as deprecated

taysea:
#### What does this PR do?
Marks `disabled-text` as deprecated and instructs users to use `text-weak` instead. Updates value of `disabled-text` to `text-weak` just to be consistent. Updates 2 instances of disabled border to use `border-weak` instead of `text-weak` just for semantic alignment.

Tested locally in DS site.

Closes https://github.com/grommet/hpe-design-system/issues/1102

Backwards compatible.

`disabled-text` is deprecated. Use `text-weak` instead.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -64,6 +64,7 @@ export const hpe = deepFreeze({
         light: '#BBBBBB',
       },
       'text-xweak': 'text-weak',
+      'text-disabled': 'text-weak',
       border: {
         dark: '#7887A1',
         light: '#999999',
@@ -79,10 +80,7 @@ export const hpe = deepFreeze({
       control: 'green',
       'active-background': 'background-contrast',
       'active-text': 'text',
-      'disabled-text': {
-        dark: '#777777',
-        light: '#999999',
-      },
+      'disabled-text': 'text-disabled', // keeping for backwards compatibility, should deprecate in next release
       'selected-background': 'green',
       'selected-text': 'text-strong',
       'status-critical': {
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -64,6 +64,7 @@ export const hpe = deepFreeze({
         light: '#BBBBBB',
       },
       'text-xweak': 'text-weak',
+      'text-disabled': 'text-weak',
```
ShimiSun, line 67: What is the long-term purpose to great a new color but use an existing color for it? Is there a plan to adjust `text-weak`?

ericsoderberghp: If this is specific to disabled input elements, I think it should be done via `theme.global.input.disabled.color`.

taysea: There is no plan to adjust text-weak as of what I know, so that is a good point about why we would need to introduce a new color name. Let me follow up with Greg on the motivation here, but maybe we do not need to do this and can just mark `disabled-text` as deprecated and instruct people to use `text-weak`

taysea: A follow-up here after discussion with Greg -- the intent wasn't that we needed to introduce a new color but instead just clean up color usage where possible. I will update this PR to just add a comment to `disabled-text` that it is deprecated and instead `text-weak` should be used. It will mirror what was done here for `status-disabled`: https://github.com/grommet/grommet-theme-hpe/pull/138

ShimiSun (approved):

ShimiSun:
Can we link to the PRs that are removing the deprecated colors from the site?

taysea:
> Can we link to the PRs that are removing the deprecated colors from the site?

Yes, I have linked it above. `disabled-text` was never referenced on the site but I updated with regards to `status-disabled`.

PR #134, opened 2020-11-03
Set default input icon color

taysea:
#### What does this PR do?
Inputs call for an icon of color `text-strong` by default. This sets that value in the theme so users don't need to set it on each implementation.

Design example can be seen on DateInput: https://www.figma.com/file/Mf9if6zyGETTWW5b9PsYNQ/HPE-Date-Input-Component?node-id=0%3A1

Test locally in design system site.

Screen Shot 2020-11-02 at 5 27 20 PM
Screen Shot 2020-11-02 at 5 27 13 PM

New visual, but not breaking.

By default, icons used in inputs now use `text-strong` instead of `text`.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -739,6 +739,20 @@ export const hpe = deepFreeze({
       border: undefined,
     },
   },
+  maskedInput: {
+    container: {
+      extend: ({ theme }) => `
+        svg {
+          fill: ${
+            theme.global.colors['text-strong'][theme.dark ? 'dark' : 'light']
+          };
+          stroke: ${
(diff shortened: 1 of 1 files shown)


halocline (approved):
Looks good to me.

ShimiSun:
Can you please elaborate on:
>Is this change backward compatible or could it be a breaking change for the official HPE theme?
New visual, but not breaking.

What will be the exact change and would it apply out of the box or only when the users will be using an icon for either MaskedInput or TextInput?

ShimiSun (approved):

PR #140, opened 2020-11-17
Modify FormField margins to align with Figma

halocline:
#### What does this PR do?

Modifies FormField margins to align with Figma by adjusting where margins are applied.

[Figma](https://www.figma.com/file/3fkwBelW5UsCbfwdDCJkT8/HPE-Form-Templates?node-id=54%3A14) calls for 12px of spacing between form fields. However, this spacing was too large in certain circumstances (18px instead of the desired 12px).

The old way:
- formField label, help, error, info had zero margin
- `formField.content.margin.vertical` = 6px
- `formField.margin.bottom` = 12px

This resulted in scenarios where if there were no error message, the spacing between that FormField and the following would have 18px of spacing in between (6px from `formField.content.margin.bottom` + 12px from `formField.margin.bottom`)

The new way:
- `formField.label.margin.top` = 6px
- `formField.error.margin.bottom` = 6px
- `formField.content.margin.vertical` = 6px

This approach ensures that regardless of whether a label or error message are present or not, the configuration of the FormField will always have 6px top and bottom margins and the desired 12px of spacing between fields.

Design System site, plus [Applitools preview deploy](https://eyes.applitools.com/app/test-results/00000251796652620064/?accountId=diU-zN8vi0mdpzHhZndu1A~~).

https://github.com/grommet/hpe-design-system/issues/1256

Diff shows expected shift in new vs. baseline, and results in correct spacing of 12px instead of 18px:
Screen Shot 2020-11-17 at 2 45 04 PM
Screen Shot 2020-11-17 at 2 45 14 PM

Backwards compatible, but will result in a visual regression of 6px shifts between form fields.

Modified spacing between FormFields in theme to align with Figma.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -571,7 +571,11 @@ export const hpe = deepFreeze({
       icon: <CircleAlert size="small" style={{ marginTop: '4px' }} />,
       size: 'xsmall',
       color: 'text',
-      margin: 'none',
+      margin: {
+        bottom: 'xsmall',
+        top: 'none',
+        horizontal: 'none',
+      },
     },
     focus: {
       border: {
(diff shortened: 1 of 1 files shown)


taysea (approved):
This implementation looks fine to me. We will need to ensure that the Figma files in all instances are updated to match this (if they do not already).

ShimiSun (approved):

PR #156, opened 2021-01-27
Use text-strong for headings

### LINKED ISSUE #152: Update grommet-theme-hpe to use `text-strong` for Headings by default

taysea:
#### What does this PR do?

Updates heading to use `text-strong` instead of `text`.

Tested locally in design system site first. See snapshots below.

A design decision was made to use `text-strong` for headings by default.

Closes #152 

Pulled these as some examples, for all snapshot tests, see: https://eyes.applitools.com/app/test-results/00000251790534341231?accountId=diU-zN8vi0mdpzHhZndu1A~~&display=details&top=00000251790534341231%2830%29

Screen Shot 2021-01-27 at 9 16 39 AM

Before:
Screen Shot 2021-01-27 at 9 16 20 AM

After:
Screen Shot 2021-01-27 at 9 16 26 AM

New visual style.

Heading uses `text-strong` by default.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -700,6 +700,7 @@ export const hpe = deepFreeze({
     round: '4px',
   },
   heading: {
+    color: 'text-strong',
     level: {
       '1': {
         small: {
@@ -835,10 +836,6 @@ export const hpe = deepFreeze({
       },
     },
     weight: 700,
-    extend: ({ theme }) =>
-      `color: ${
-        theme.global.colors['text-strong'][theme.dark ? 'dark' : 'light']
-      };`,
   },
   icon: {
     size: {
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -835,6 +835,10 @@ export const hpe = deepFreeze({
       },
     },
     weight: 700,
+    extend: ({ theme }) =>
```
ericsoderberghp, line 838: darn, I was hoping we could do this via something like `theme.heading.color`, similar to how we have `theme.heading.weight`.

taysea: I think that would be a good enhancement. Grommet doesn't currently support that, so I was going with what functionality already existed, but I can file a PR on grommet for this.

taysea: I've incorporated the new grommet changes.

ericsoderberghp (approved):
So simple :)

PR #159, opened 2021-02-01
Update tab styling for active state and padding

### LINKED ISSUE #146: Tabs highlighting isn't prominent enough for the users to identify which tab is selected.
- [ ] Ideate a few other approaches
- [ ] run some A/B UserZoom tests

taysea:
#### What does this PR do?
Updates tab styling for active state and increases horizontal padding to create better affinity between tab icon + label.

Tested locally in aries-site.

Closes #146 
Closes https://github.com/grommet/hpe-design-system/issues/1363
Closes https://github.com/grommet/hpe-design-system/issues/1385

Screen Shot 2021-02-01 at 9 39 45 AM

Introduces new tab styling.

Active tabs have a "medium" width border of border-strong and increased horizontal padding from "small" to "medium".

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1015,10 +1015,10 @@ export const hpe = deepFreeze({
         color: 'border-strong',
       },
       disabled: {
-        color: 'transparent',
+        color: undefined,
       },
       hover: {
-        color: 'transparent',
+        color: undefined,
       },
     },
     disabled: {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1009,21 +1009,28 @@ export const hpe = deepFreeze({
     },
     border: {
       side: 'bottom',
-      color: 'border',
+      color: 'transparent',
+      size: 'medium',
       active: {
         color: 'border-strong',
       },
       disabled: {
-        color: 'border-weak',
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -1009,21 +1009,28 @@ export const hpe = deepFreeze({
     },
     border: {
       side: 'bottom',
-      color: 'border',
+      color: 'transparent',
+      size: 'medium',
       active: {
         color: 'border-strong',
       },
       disabled: {
```
ericsoderberghp, line 1017: Would it work to do `disabled: undefined` or `disabled: {}` or `disabled: { color: undefined }` instead? Just trying to keep the DOM cleaner where possible.

taysea: I will double check but I believe we needed transparent to maintain the proper dimensions / text alignment. let me try with undefined now though.

taysea: I've done `disabled: { color: undefined }` in the site locally, and that works.

taysea: same for hover.

ericsoderberghp (approved):

PR #167, opened 2021-02-22
add tooltip

britt6612:
Add the tooltip theme to match the designs.

Currently, the designs call for  `medium` elevation. However, it does not look correct with the medium so  I replaced it with small and will follow up on an issue. 

With medium, the bottom shadow is very harsh. 
Screen Shot 2021-02-22 at 11 50 34 AM

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1123,6 +1123,7 @@ export const hpe = deepFreeze({
       border: {
         color: 'border-weak',
       },
+      margin: 'xxsmall',
       elevation: 'small',
       pad: {
         vertical: 'none',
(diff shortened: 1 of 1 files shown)


taysea (approved):
LGTM. Thanks for noting the issue with elevation getting cut off. I notice the harsh edge where it seems like the bounds of the elevation is being restricted. Can you link to the grommet issue once it's filed?

@ericsoderberghp are we comfortable using `small` for the elevation for now until that elevation cut-off issue is resolved? designs are calling for `medium`

halocline (approved):

Issue #160, opened 2021-02-03
npm install grommet-theme-hpe --save throws error


sathishiva:
**Version from my machine:**
node version: 15.0.1
npm version: 7.0.3
react-scripts: 4.0.2

**Log:**
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: ccs-ui-components@0.1.0
npm ERR! Found: react@17.0.1
npm ERR! node_modules/react
npm ERR!   react@"^17.0.1" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.13.1" from grommet-theme-hpe@2.0.2
npm ERR! node_modules/grommet-theme-hpe
npm ERR!   grommet-theme-hpe@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!

halocline:
Can you try `npm install --legacy-peer-deps` as suggested 3rd line from bottom of the log?

ShimiSun:
Adding to @halocline, you might find the generic description of the problem and the way to troubleshoot it on https://stackoverflow.com/questions/64573177/unable-to-resolve-dependency-tree-error-when-installing-npm-packages

halocline:
Yes, my apologies for not providing additional detail for reference.  NPM v7 made the decision to "always run with --legacy-peer-deps set to false, so that missing peer dependencies will be highlighted when looking at the package tree." This differed from how NPM handled v4-6. Here is [NPM's blog post](https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major) discussing v7 changes.

sathishiva:
Thanks @ShimiSun @halocline for the quick and detailed explanation!
It works with `npm install grommet-theme-hpe  --legacy-peer-deps`

sathishiva:
If we have to run the `npm install` in CI. This won't work. Noticed that hpe package JSON using react version with ^. It has to be similar to grommet package.json (using     "react": ">= 16.6.1")
https://github.com/grommet/grommet/blob/master/package.json#L47
https://github.com/grommet/grommet-theme-hpe/blob/master/package.json#L25

mukarramgit:
hi

[...]

PR #177, opened 2021-03-24
update react version

britt6612:
#### What does this PR do?
This PR updates react version
I tested this with `npm link` and then the dist folder was copied into a create react app that I had running locally. I ran `npm install` and got no error.

We can always merge this to another branch besides master to be able to fully test the `npm install grommet-theme-hpe` since I  could only run `npm install` after the dist folder was pasted into my node_modules.

Starting from npm 7, peerDependencies are automatically installed together with normal dependencies. When npm finds a version mismatch, a problem arises because it will abort the installation. grommet-theme-hpe has react dependency of react: 16.13.1. However, if someone is using React 17 or above, a mismatch will be detected and the installation will be aborted.

Recommended solution is to change the grommet-theme-hpe dependencies to match that of grommet:
 "react": "^16.13.1 || ^17.0.0",

Other workarounds on the implementer side would be one of the following (however this may be less desirable):
-Downgrade react version on the project
-Downgrade npm
-Ignore the peer dependency by using npm install `grommet-theme-hpe` --legacy-peer-deps

closes issue #160 
 backward compatible

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "peerDependencies": {
     "grommet": "^2.16.3",
     "grommet-icons": "^4.5.0",
-    "react": "^16.13.1",
+    "react": "^16.13.1 || ^17.0.1",
     "styled-components": "^5.0.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,9 +3,9 @@
 
 
 "@babel/cli@^7.1.2":
-  version "7.12.10"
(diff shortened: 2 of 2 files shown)


taysea (approved):
Thanks for the explanation. This change makes sense to me to align with grommet peerDependency. Will be good to test in a sample app with `npm` once merged.

ShimiSun (approved):
Nice work!

PR #180, opened 2021-04-08
clean up drop margin

britt6612:
use new intelligent margin from grommet

#### What does this PR do?
Adds `intelligent margin` to theme both `search` and `tooltip` are `xsmall` margin and now with the new `intelligent margin` the side in which the margin is placed is accurate.
design system local theme

```
  global: {
    drop: {
      margin: 'xsmall',
      intelligentMargin: true,
      background: 'background-front',
      border: {
        radius: '4px',
      },
      extend: () => {},
    },
  },

```
https://github.com/grommet/hpe-design-system/issues/1449
backward compatible

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -223,10 +223,8 @@ export const hpe = deepFreeze({
       border: {
         radius: '4px',
       },
-      extend: ({ alignProp, theme }) => `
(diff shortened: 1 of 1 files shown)


taysea (approved):
LGTM. tested locally on site as well

ShimiSun (approved):

Issue #187, opened 2021-04-30
Drop content (Select options) does not appear when Layer has elevated zIndex


halocline:
After increasing Layer's zIndex, Drop no longer appears when in a Layer:
![Screen Shot 2021-04-30 at 9 22 30 AM](https://user-images.githubusercontent.com/1756948/116717139-d09b6c00-a995-11eb-9b76-32408d8ed655.png)

Setting Drop's zIndex to be equivalent with Layer renders desired results:
![Screen Shot 2021-04-30 at 9 22 38 AM](https://user-images.githubusercontent.com/1756948/116717292-f6287580-a995-11eb-8456-09472840c1e6.png)

PR #185, opened 2021-04-28
Adjust Layer z-index to sit atop HPE global header

halocline:
#### What does this PR do?

Closes https://github.com/grommet/grommet-theme-hpe/issues/184

#### What testing has been done on this PR?

Local project implementation with the Global header service.

#### Any background context you want to provide?

https://github.com/grommet/grommet-theme-hpe/issues/184

#### What are the relevant issues?

https://github.com/grommet/grommet-theme-hpe/issues/184

#### Is this change backward compatible or could it be a breaking change for the official HPE theme?

bc

Issue #184, opened 2021-04-27
Resolve Conflict - HPE Global Header Footer Service & Layer z-indexes conflict


halocline:
The new HPE Global Header/Footer service has a z-index of 101. Layer's default z-index is 20.

- Should a layer such as a modal or side drawer sit on top of the global header/footer?
- If yes, should we update HPE theme to set Layer's z-index to a value greater than 101?

**Without adjustment**
![Screen Shot 2021-04-27 at 8 51 49 AM](https://user-images.githubusercontent.com/1756948/116262670-f75b6780-a735-11eb-8cba-95a2e6266de5.png)

**After adjustment**
![Screen Shot 2021-04-27 at 8 52 13 AM](https://user-images.githubusercontent.com/1756948/116262697-fb878500-a735-11eb-9bac-dd6030010cde.png)

PR #188, opened 2021-04-30
Align Drop and Layer zIndex

halocline:
#### What does this PR do?

Closes https://github.com/grommet/grommet-theme-hpe/issues/187

Tested locally with `<Select />` within a `<Layer />`. However, I think we should do more exhaustive testing to check for edge cases and different combinations of components presented in a Layer.

I could use help brainstorming any other potential areas which might have unintended side effects do to the increased zIndex.

Since we are trying to release the theme, we may want to revert https://github.com/grommet/grommet-theme-hpe/pull/185, then use this implementation in another project to test on the side over an extended period of time.

Originally trying to solve for the HPE Global Header Service which has a `z-index` of 101 and prevents Layer from filling window properly.

 https://github.com/grommet/grommet-theme-hpe/issues/184
https://github.com/grommet/grommet-theme-hpe/issues/187

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -226,6 +226,10 @@ export const hpe = deepFreeze({
       margin: 'xsmall',
       intelligentMargin: true,
       shadowSize: 'medium',
+      /* HPE Global Header/Footer Service a.k.a. HPE Common HFWS sets the header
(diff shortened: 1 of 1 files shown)


ericsoderberghp (approved):

taysea (commented):
Thanks for catching this.

PR #192, opened 2021-05-25
Add toolbar button kind

taysea:
#### What does this PR do?
This adds a button kind `toolbar` for the button kind featured below. Currently, this kind of button is always used in a toolbar context where it is inline with inputs such as a Search and the desire is that it has equal prominence as its peers.

Test in DS site.

Closes https://github.com/grommet/hpe-design-system/issues/1118

Screen Shot 2021-05-25 at 10 58 45 AM

Backwards compatible.

Added `toolbar` button kind.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -313,6 +313,16 @@ export const hpe = deepFreeze({
         weight: 700,
       },
     },
+    toolbar: {
+      border: {
+        color: 'border',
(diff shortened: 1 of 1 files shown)


ericsoderberghp (approved):
Wow, gotta like how easy this is.

halocline (approved):

PR #196, opened 2021-06-22
Update CheckBox theming from latest grommet enhancements

taysea:
#### What does this PR do?
Now that we have [enhanced Grommet](https://github.com/grommet/grommet/pull/5288) to better handle focus behavior of checkboxes, we can remove this from the extend. Previously, we had been removing the border since most CheckBoxes are displayed within FormFields, but this was causing accessibility issues in cases where CheckBox is outside of FormField. We no longer have this conflict.

Tested in DS site. Checkbox outside of FormField receives focus on check. Checkbox inside formfield receives hover treatment.

Backwards compatible.

Fixed CheckBox focus behavior to have focus outline when outside FormField.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -480,7 +480,6 @@ export const hpe = deepFreeze({
     check: {
       radius: '2px',
       extend: ({ theme, checked, indeterminate }) => `
-      box-shadow: none;
       background-color: ${
         checked || indeterminate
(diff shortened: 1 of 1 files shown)


ericsoderberghp (approved):

ShimiSun:
We'll need to update peerdependency of grommet for the release that will contain this fix. @taysea

Issue #200, opened 2021-07-28
DataTable Header blur not supported by Firefox


taysea:
Firefox doesn't support `backdrop-filter` in CSS, so the blur effect that we have behind DataTable headers on scroll doesn't come through. Is this something we should be concerned about in terms of readability of the table headers? 

On Firefox:
<img width="1297" alt="Screen Shot 2021-07-28 at 2 20 48 PM" src="https://user-images.githubusercontent.com/12522275/127397863-8ed7432d-9454-40e8-8ed1-c6dd006ffbf6.png">

On Chrome:
<img width="1295" alt="Screen Shot 2021-07-28 at 2 23 24 PM" src="https://user-images.githubusercontent.com/12522275/127397982-3dbbacad-3bdd-452f-8848-ce9a0eca7067.png">

PR #202, opened 2021-08-05
add alternative background for Firefox 

britt6612:
#### What does this PR do?
This PR adds a background that has a opacity for `datatable` for just FireFox browser
Browser
We were using  `backdrop-filter: blur(8px)` however FireFox does not support `backdrop-filter:`
so as an alternative I thought adding a background will at least make the `datatable` more readable 
closes issue #200 
Currently looks like this below: 
Screen Shot 2021-08-04 at 9 24 09 PM
Screen Shot 2021-08-04 at 9 24 22 PM

With the background:
Does not have that nice effect as chrome but at least now its much easier to read:

Screen Shot 2021-08-04 at 9 25 43 PM
Screen Shot 2021-08-04 at 9 25 58 PM

In Chrome:

Screen Shot 2021-08-04 at 9 26 56 PM

backward compatible
sure

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -602,21 +602,30 @@ export const hpe = deepFreeze({
           color: 'background-front',
           opacity: 'strong',
         },
-        extend: 'backdrop-filter: blur(8px);',
+        extend: `backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+          background-color: rgba(255, 255, 255, 0.92);
+      }`,
       },
       body: {
         background: {
           color: 'background-front',
           opacity: 'strong',
         },
-        extend: 'backdrop-filter: blur(8px);',
+        extend: `backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+          background-color: rgba(255, 255, 255, 0.92);
+      }`,
       },
       footer: {
         background: {
           color: 'background-front',
           opacity: 'strong',
         },
-        extend: 'backdrop-filter: blur(8px);',
+        extend: `backdrop-filter: blur(8px);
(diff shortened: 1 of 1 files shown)


halocline:
@britt6612 How does this treatment work with the dark theme? Is there a demo branch you can create on the Design System site so that we can see the applied treatments?

britt6612:
> @britt6612 How does this treatment work with the dark theme? Is there a demo branch you can create on the Design System site so that we can see the applied treatments?

I will create a PR in design system and link it here!

britt6612:
changes can be seen here https://deploy-preview-1789--keen-mayer-a86c8b.netlify.app/components/table on FireFox

src/js/index.js
```diff
@@ -596,27 +596,42 @@ export const hpe = deepFreeze({
...
         background: {
           color: 'background-front',
           opacity: 'strong',
         },
-        extend: 'backdrop-filter: blur(8px);',
+        extend: ({ theme }) => `backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+          background: ${theme.dark ? '#404b5cF2' : '#ffffffF2'};
+          }
+      }`,
```
halocline, line 612: just a little reminder for where these colors came from

```suggestion
        // values for background are same as 'background-front' set to a 95% opacity
        extend: ({ theme }) => `backdrop-filter: blur(8px);
        @-moz-document url-prefix() {
          background: ${theme.dark ? '#404b5cF2' : '#ffffffF2'};
          }
      }`,
```

taysea (approved):
Nice job on this.

ericsoderberghp (approved):

PR #208, opened 2021-09-08
change the padding to take into account the border

britt6612:
#### What does this PR do?
This PR changes the padding for the `input` this takes into account the border for the inputs. The size of the `textinput` without being wrapped in a `formfield` is `38px` with wrapping it in a `formfield` then it uses the `textinput` plain so without the border and is then the correct size.
in the design system site
Below is a picture of the sizes for `button` the padding is off by one to account for the border. I believe we need to do the same for the `inputs` as well. Since `Select` and  `Menu` are made from `button` their sizes are correct. 
Screen Shot 2021-09-08 at 3 59 25 PM

With this change to the padding you can see that they both will match up correctly. 
Screen Shot 2021-09-08 at 4 02 00 PM

Closes https://github.com/grommet/hpe-design-system/issues/1778
backward compatible

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -164,8 +164,8 @@ export const hpe = deepFreeze({
         weight: 500,
       },
       padding: {
-        horizontal: 'small',
-        vertical: 'xsmall',
+        horizontal: '11px',
+        vertical: '5px',
       },
       extend: `
         &::-webkit-input-placeholder {
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -164,7 +164,7 @@ export const hpe = deepFreeze({
         weight: 500,
       },
       padding: {
-        horizontal: '11px',
+        horizontal: '11px', // equivalent to 'small' when combined with 1px border
         vertical: '5px', // equivalent to 'small' when combined with 1px border
       },
       extend: `
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -165,7 +165,7 @@ export const hpe = deepFreeze({
       },
       padding: {
         horizontal: '11px', // equivalent to 'small' when combined with 1px border
-        vertical: '5px', // equivalent to 'small' when combined with 1px border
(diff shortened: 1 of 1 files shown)


src/js/index.js
```diff
@@ -164,8 +164,8 @@ export const hpe = deepFreeze({
         weight: 500,
       },
       padding: {
-        horizontal: 'small',
-        vertical: 'xsmall',
+        horizontal: '11px',
```
taysea, line 167:
```suggestion
        horizontal: '11px', // equivalent to 'small' when combined with 1px border
```

src/js/index.js
```diff
@@ -164,8 +164,8 @@ export const hpe = deepFreeze({
...
+        vertical: '5px',
```
taysea, line 168:
```suggestion
        vertical: '5px', // equivalent to 'small' when combined with 1px border
```

src/js/index.js
```diff
@@ -164,8 +164,8 @@ export const hpe = deepFreeze({
         weight: 500,
       },
       padding: {
-        horizontal: 'small',
-        vertical: 'xsmall',
+        horizontal: '11px',
+        vertical: '5px', // equivalent to 'small' when combined with 1px border
```
taysea, line 168:
```suggestion
        vertical: '5px', // equivalent to 'xsmall' when combined with 1px border
```

taysea (approved):

taysea (commented):
Thinking we should put a PR up on design-system-site just to check if any snapshot changes exist that we might not have caught in manual testing, but otherwise looks good.

britt6612:
> Thinking we should put a PR up on design-system-site just to check if any snapshot changes exist that we might not have caught in manual testing, but otherwise looks good.

oh yes good thought Ill do that!

taysea:
Theme changes on local deployment of site look good. Good to go here!

ericsoderberghp (approved):

PR #210, opened 2021-09-15
Update pinned DataTable background

### LINKED ISSUE #206: DataTable header and primary column backgrounds are incorrect in darkmode
See image... in darkmode, DataTable background shading is incorrect/broken. This then makes things odd for row-hover etc.

Screenshot is taken from current HPE DS website:
Screen Shot 2021-08-26 at 10 41 39 AM

taysea:
#### What does this PR do?

Leverages Grommet's more recent intelligent background for datatable pinned cells. Thus we can remove the "background-front" declaration and let the datatable be smart about what background it's on.

Updating the opacity from 'strong' to 0.95 so it works for all browsers (Firefox included). This does make the blur slightly less apparent in Chrome but I think the difference is minimal and creates a more readable table.

Tested locally in DS Site.

Closes #206 

Note: these are static images, the blur is more visible when user is actually scrolling.

Screen Shot 2021-09-15 at 1 54 27 PM
Screen Shot 2021-09-15 at 1 54 05 PM

Makes the opacity on the pinned cells from 0.8 --> 0.95 but fixes incorrect color treatment/makes behavior more consistent from firefox to chrome and other browsers.

- Fixed pinned DataTable background color

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -12,12 +12,12 @@ import { FormUp } from 'grommet-icons/icons/FormUp';
 import { FormTrash } from 'grommet-icons/icons/FormTrash';
 import { Unsorted } from 'grommet-icons/icons/Unsorted';
 
-const isObject = item =>
+const isObject = (item) =>
   item && typeof item === 'object' && !Array.isArray(item);
 
-const deepFreeze = obj => {
+const deepFreeze = (obj) => {
   Object.keys(obj).forEach(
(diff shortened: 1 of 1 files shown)


britt6612 (changes requested):
This looks good code-wise! Not sure about the eslint changes ?

britt6612 (approved):

ericsoderberghp (approved):

PR #214, opened 2021-09-28
Added dark and light code block themes

jcfilben:
#### What does this PR do?
Adds the code block dark and light themes

Related to https://github.com/grommet/hpe-design-system/pull/1903

Backwards compatible

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1233,3 +1233,253 @@ export const hpe = deepFreeze({
 });
 
 export const { colors } = hpe.global;
+
+export const DarkCodeTheme = {
+  'code[class*="language-"]': {
+      'color': colors.text.dark,
+      'background': 'none',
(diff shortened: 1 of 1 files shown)


halocline:
I've created https://github.com/grommet/grommet-theme-hpe/pull/215 as an alternate structure for consideration.

src/js/index.js
```diff
@@ -1233,3 +1233,253 @@ export const hpe = deepFreeze({
 });
 
 export const { colors } = hpe.global;
+
+export const DarkCodeTheme = {
```
taysea, line 1237: Just a general naming thing, should we do camel case? `darkCodeTheme` and `lightCodeTheme`?

jcfilben: Yes I will update the names to be camel case

PR #215, opened 2021-10-04
Add code theme as export

halocline:
#### What does this PR do?

For consideration, an alternate structure/approach to https://github.com/grommet/grommet-theme-hpe/pull/214. 

Proposes consumption of theme objects like:
```
import { hpe, code } from 'grommet-theme-hpe';
 ```

Applied `/dist` build to local HPE Design System instance.

https://github.com/grommet/hpe-design-system/issues/1870

Backwards

Added `code` theme object to enable hpe-themed code syntax highlighting.

### CODE REVIEW

**Comment by halocline** [comment]
Tagging @jcfilben for review.

**Review by britt6612 — APPROVED** [review]
Nice work I think this is a lot cleaner and having the code separate from the hpe keeps things very organized.

**Review by jcfilben — APPROVED** [review]
I like this structure!

File: src/js/themes/index.d.ts [hunk 1]
```diff
@@ -0,0 +1,5 @@
+declare const hpe: object;
+declare const colors: object;
+declare const codeTheme: object;
+
+export { hpe };
```
taysea, line 5:
```suggestion
export { hpe };

```

src/js/index.d.ts
```diff
@@ -1,3 +1 @@
-declare const hpe: object;
-
-export { hpe };
\ No newline at end of file
+export * from './themes';
```
taysea, line 1:
```suggestion
export * from './themes';

```

taysea (commented):
Looks good, just added new line to a few files that had been flagged!

src/js/themes/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare const hpe: object;
+declare const colors: object;
+declare const codeTheme: object;
```
ericsoderberghp, line 3: Should this be "code" instead of "codeTheme"?

src/js/themes/code.js
```diff
@@ -0,0 +1,173 @@
...
+  },
+  bold: {
+    fontWeight: 'bold',
+  },
+  italic: {
+    fontStyle: 'italic',
+  },
+});
+
+export const code = {
```
ericsoderberghp, line 170: Is this schema specific to a particular syntax highlighter or general for any? If specific, I wonder if we should be a little less generic in the name.

jcfilben: It is specific to Prism

halocline: Good call. Will do, plus we can add a comment which references the schema documentation.

PR #222, opened 2021-11-09
align label to be start

britt6612:
#### What does this PR do?
This PR aligns label for `checkbox` so when there is multi-lines it is aligned start
tested with aries.theme in design-system
currently we have:
Screen Shot 2021-11-09 at 12 44 44 PM

what we want:
Screen Shot 2021-11-09 at 12 45 12 PM

closes https://github.com/grommet/hpe-design-system/issues/2000
look above
no
yes

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -14,12 +14,12 @@ import { Unsorted } from 'grommet-icons/icons/Unsorted';
 
 import { colors } from './colors';
 
-const isObject = item =>
+const isObject = (item) =>
   item && typeof item === 'object' && !Array.isArray(item);
 
-const deepFreeze = obj => {
+const deepFreeze = (obj) => {
   Object.keys(obj).forEach(
-    key => key && isObject(obj[key]) && Object.freeze(obj[key]),
+    (key) => key && isObject(obj[key]) && Object.freeze(obj[key]),
   );
   return Object.freeze(obj);
 };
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -14,12 +14,12 @@ import { Unsorted } from 'grommet-icons/icons/Unsorted';
 
 import { colors } from './colors';
 
-const isObject = (item) =>
+const isObject = item =>
```
taysea, line 17: seems like we lost some prettier changes that we want

taysea (commented):
just need to check out some linting changes -- seems like it might not be on latest prettier. but the theme stuff looks good

src/js/themes/hpe.js
britt6612: fixing now

taysea (approved):
Thanks!

ericsoderberghp:
I'm concerned that when a custom element is used for the `label` that it won't always align well. For example, if the caller does `label={<Text size="large">my fancy label</Text>}` then the box will be aligned oddly. I think we might need to add a property to CheckBox to allow the caller to control this based on what the label might be.

britt6612:
> I'm concerned that when a custom element is used for the `label` that it won't always align well. For example, if the caller does `label={<Text size="large">my fancy label</Text>}` then the box will be aligned oddly. I think we might need to add a property to CheckBox to allow the caller to control this based on what the label might be.

If users are using the Design-System I don't believe they would be passing something like 
`label={<Text size="large">my fancy label</Text>}`  but this is just my opinion that it would not be under our recommended `labels` we can still explore this enhancement on grommet side

britt6612:
> I'm concerned that when a custom element is used for the `label` that it won't always align well. For example, if the caller does `label={<Text size="large">my fancy label</Text>}` then the box will be aligned oddly. I think we might need to add a property to CheckBox to allow the caller to control this based on what the label might be.

@ericsoderberghp with the new Grommet change this will take care of the "fancy" label concern 
`label={<Text size="large">my fancy label</Text>}`

ericsoderberghp (approved):

PR #223, opened 2021-12-15
add hover for primary button

britt6612:
#### What does this PR do?
this PR adds new hover behavior for `button`
closes https://github.com/grommet/hpe-design-system/issues/2122
backward compatible
yes

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -254,7 +254,7 @@ export const hpe = deepFreeze({
       primary: {
         // Designs call for using text-strong for the text color.
         // Keeping the label color the same on hover to maintain readablility.
-       color: 'text-strong',
+        color: 'text-strong',
         background: {
           dark: 'validation-ok',
           // Designs call for using a lower opacity of exisiting green!
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -251,6 +251,33 @@ export const hpe = deepFreeze({
         },
         color: undefined,
       },
+      extend: ({ colorValue, primary, theme }) => `
+      ${
+        primary &&
+        colorValue &&
+        `:hover {
+        background-color: ${
+          theme.dark
+            ? theme.global.colors[colorValue].dark
+            : theme.global.colors[colorValue].light
+        };
+      }`
+      }
+    `,
+      primary: {
+        // Designs call for using text-strong for the text color.
+        // Keeping the label color the same on hover to maintain readablility.
+        // color: hpe.global.colors['text-strong'].light,
+        background: {
+          dark: 'validation-ok',
+          // Designs call for using a lower opacity of exisiting green!
+          // instead of creating a new color name for this single usecase.
+          light: {
+            color: 'green!',
+            opacity: 0.75,
+          },
+        },
+      },
       secondary: {
         border: {
           width: '3px',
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -251,6 +251,33 @@ export const hpe = deepFreeze({
         },
         color: undefined,
       },
+      extend: ({ colorValue, primary, theme }) => `
```
taysea, line 254: Can we put extend at the end of this block just to minimize emphasis on extend?

britt6612: yup!

src/js/themes/hpe.js
```diff
@@ -266,6 +280,19 @@ export const hpe = deepFreeze({
...
```
ericsoderberghp, line 283: Is there a reason we cannot use `theme.button.hover.primary.background` for this, with `dark` and `light` sub-keys?

britt6612: so the reason I have this is to keep backward compatibility until `colored` button is discussed and researched further. Since I have the above `button.hover.primary.background` set to the dark green above this would also change the `colored` button since we pass primary color here: 
```
      primary: {
        // Designs call for using text-strong for the text color.
        // Keeping the label color the same on hover to maintain readablility.
        // color: hpe.global.colors['text-strong'].light,
        background: {
          dark: 'validation-ok',
          // Designs call for using a lower opacity of exisiting green!
          // instead of creating a new color name for this single usecase.
          light: {
            color: 'green!',
            opacity: 0.75,
          },
        },
      },
```
So in order to keep things the way they are I added this extend for `Primary && colorValue` are used
Im thinking we can enhance grommet to also have something such as `theme.button.hover.color.primary.baackground` then we can set a dark and light in there. I was holding off on this until the research for hover on colored buttons was complete.

src/js/themes/hpe.js
```diff
@@ -251,6 +251,20 @@ export const hpe = deepFreeze({
         },
         color: undefined,
       },
+      primary: {
+        // Designs call for using text-strong for the text color.
+        // Keeping the label color the same on hover to maintain readablility.
+        // color: hpe.global.colors['text-strong'].light,
```
taysea, line 257:
```suggestion
        color: hpe.global.colors['text-strong'].light,
```

taysea, line 257: Think we still need to apply this label color, but we need to check how to get text-strong.light since we don't have access to `hpe` here I don't think.

taysea, line 257:
```suggestion
       color: 'text-strong',
```
In both light and dark mode, we want to ensure the label color stays as text-strong on hover

britt6612: your right Ill fix

src/js/themes/hpe.js
```diff
@@ -251,6 +251,20 @@ export const hpe = deepFreeze({
         },
         color: undefined,
       },
+      primary: {
+        // Designs call for using text-strong for the text color.
+        // Keeping the label color the same on hover to maintain readablility.
+       color: 'text-strong',
```
taysea, line 257:
```suggestion
        color: 'text-strong',
```

taysea (approved):

ericsoderberghp (approved):

PR #226, opened 2021-12-20
fix label color for button hover on colored button

britt6612:
#### What does this PR do?
This PR fixes the `label` color from changing for `colored` buttons
yarn link into a grommet project 
backward compatible

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "eslint-plugin-prettier": "^3.0.0",
     "eslint-plugin-react": "^7.19.0",
     "fs-extra": "^10.0.0",
+    "grommet": "^2.19.1",
     "grommet-icons": "^4.4.0",
     "pre-commit": "^1.2.2",
     "prettier": "^2.3.2",
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -1,5 +1,6 @@
 // (C) Copyright 2020-2021 Hewlett Packard Enterprise Development LP
 import React from 'react';
+import { backgroundIsDark } from 'grommet/utils';
 import { css } from 'styled-components';
 import { Ascending } from 'grommet-icons/icons/Ascending';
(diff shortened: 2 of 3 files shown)


src/js/themes/hpe.js
```diff
@@ -285,10 +286,17 @@ export const hpe = deepFreeze({
...
+          backgroundIsDark(colorValue, theme)
+            ? theme.global.colors['text-strong'].dark
+            : theme.global.colors['text-strong'].light
+        };
         background-color: ${
           theme.dark
-            ? theme.global.colors[colorValue].dark
-            : theme.global.colors[colorValue].light
+            ? theme.global.colors[colorValue].dark ||
+              theme.global.colors[colorValue]
```
britt6612, line 297: had to account for if a user passes in `purple!` that doesn't have a `light` or `dark`

package.json
```diff
@@ -44,6 +44,7 @@
     "eslint-plugin-prettier": "^3.0.0",
     "eslint-plugin-react": "^7.19.0",
     "fs-extra": "^10.0.0",
+    "grommet": "^2.19.1",
```
ericsoderberghp, line 47: We've tried to avoid circular dependencies, since `grommet` depends on `grommet-theme-hpe` already.

src/js/themes/hpe.js
```diff
@@ -285,10 +286,17 @@ export const hpe = deepFreeze({
         primary &&
         colorValue &&
         `:hover {
+        color: ${
```
ericsoderberghp, line 289: This is starting to feel like too much functionality from grommet is leaking into the theme. Maybe we should revert the prior change and re-think how we want to approach this, probably by enhancing the grommet side more.

britt6612: That's what I was thinking maybe we should revert or take out the `colored` button I can make a PR on the grommet side to handle this. I agree this is heavy but didn't think of anything else.

PR #234, opened 2022-02-09
change the places we have text-weak to text-xweak

britt6612:
#### What does this PR do?
This PR goes through and changes the current values we have for `text-weak` to `text-xweak`
tested this in the design system 
 anything that was `text-weak` should be `text-xweak`
closes https://github.com/grommet/hpe-design-system/issues/2312

### DIFF
--- a/src/js/themes/colors.js
+++ b/src/js/themes/colors.js
@@ -41,7 +41,7 @@ export const colors = {
     light: '#757575',
   },
   'text-xweak': {
-    dark: '#606b7d',
+    dark: '#606B7D',
     light: '#bbbbbb',
   },
   border: {
--- a/src/js/themes/colors.js
+++ b/src/js/themes/colors.js
@@ -37,10 +37,13 @@ export const colors = {
     light: '#000000',
   },
   'text-weak': {
-    dark: '#606B7D',
-    light: '#BBBBBB',
+    dark: '#8c98aa',
+    light: '#757575',
   },
-  'text-xweak': 'text-weak',
+  'text-xweak': {
+    dark: '#606b7d',
+    light: '#bbbbbb',
+  },
   border: {
     dark: '#7887A1',
     light: '#999999',
--- a/src/js/themes/colors.js
+++ b/src/js/themes/colors.js
@@ -37,7 +37,7 @@ export const colors = {
     light: '#000000',
   },
   'text-weak': {
-    dark: '#8c98aa',
+    dark: '#8C98AA',
     light: '#757575',
   },
   'text-xweak': {
--- a/src/js/themes/colors.js
+++ b/src/js/themes/colors.js
@@ -42,7 +42,7 @@ export const colors = {
   },
   'text-xweak': {
     dark: '#606B7D',
-    light: '#bbbbbb',
+    light: '#BBBBBB',
   },
   border: {
     dark: '#7887A1',
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -230,7 +230,7 @@ export const hpe = deepFreeze({
       background: {
         color: 'transparent',
       },
-      color: 'text-weak',
+      color: 'text-xweak',
       primary: {
         border: {
           color: 'border-weak',
@@ -563,7 +563,7 @@ export const hpe = deepFreeze({
       remove: FormTrash,
     },
     message: {
-      color: 'text-weak',
+      color: 'text-xweak',
     },
     pad: { horizontal: 'xsmall' },
     extend: 'border-radius: 4px;',
@@ -588,7 +588,7 @@ export const hpe = deepFreeze({
         color: 'border-weak',
       },
       label: {
-        color: 'text-weak',
+        color: 'text-xweak',
       },
     },
     error: {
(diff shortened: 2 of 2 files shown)


britt6612:
FileInput message was changed to `text-xweak` to match other inputs such as `select` and `textinput` however this failed the WAVE test need to look into this a little more 
https://design-system.hpe.design/components/fileinput
This is also currently failing now with the `text-weak` so thinking we need another issue to dive deeper or decide if we want to bypass

ericsoderberghp:
Also, for the `hpe` theme, colors.js has `'text-xweak': 'text-weak',`, so I'm not sure how these changes would make anything different.

britt6612:
> Also, for the `hpe` theme, colors.js has `'text-xweak': 'text-weak',`, so I'm not sure how these changes would make anything different.

sorry didn't commit should be there now

src/js/themes/colors.js
```diff
@@ -37,10 +37,13 @@ export const colors = {
     light: '#000000',
   },
   'text-weak': {
-    dark: '#606B7D',
-    light: '#BBBBBB',
+    dark: '#8c98aa',
+    light: '#757575',
+  },
+  'text-xweak': {
+    dark: '#606b7d',
```
taysea, line 44: small change, but for consistency with the rest of theme can we capitalize the letters in these HEX values?

britt6612:
I commented the places that currently use `text-weak`
https://github.com/grommet/hpe-design-system/issues/2312#issuecomment-1033073576

taysea:
> I commented the places that currently use `text-weak` [grommet/hpe-design-system#2312 (comment)](https://github.com/grommet/hpe-design-system/issues/2312#issuecomment-1033073576)

Perfect, you're on top of it.

src/js/themes/colors.js
```diff
@@ -37,10 +37,13 @@ export const colors = {
     light: '#000000',
   },
   'text-weak': {
-    dark: '#606B7D',
-    light: '#BBBBBB',
+    dark: '#8c98aa',
```
taysea, line 40:
```suggestion
    dark: '#8C98AA',
```
```suggestion
    dark: '#8c98aa',
```

src/js/themes/colors.js
```diff
@@ -37,10 +37,13 @@ export const colors = {
...
+    light: '#757575',
+  },
+  'text-xweak': {
+    dark: '#606b7d',
+    light: '#bbbbbb',
```
taysea, line 45:
```suggestion
    light: '#BBBBBB',
```

src/js/themes/colors.js
```diff
@@ -37,10 +37,13 @@ export const colors = {
     light: '#000000',
   },
   'text-weak': {
+    dark: '#8C98AA',
```
MikeKingdom, line 40: Do we need to change places that talk about using this for disabled like `disabled-text` (line 62) and `status-disabled` (line 81) (not to mention I always wondered why these were deprecated :) )

britt6612: Let me look into that

taysea: Here's background:

`status-disabled` was never being used in designs, so we just deprecated it in favor of `text-weak` / `border-weak`: https://github.com/grommet/grommet-theme-hpe/pull/138
`disabled-text` was deprecated to keep more semantic alignment in the "steps" of text colors: https://github.com/grommet/grommet-theme-hpe/pull/133 -- however, I could see an advantage now that we have  a new `text-weak` to naming the new `text-xweak` as `text-disabled` as opposed to `text-xweak` so it's clear when to use that text color (since it won't meet color contrast requirements for readability). but I could go either way here.

ericsoderberghp: `text-xweak` aligns more with the grommet pattern. I could see defining `disabled-text` to be `text-xweak` to add a layer of indirection.

taysea: I'm good just leaving `text-xweak` in that case and not doing any specific disabled-text.

britt6612: @ericsoderberghp would you agree with changing `status-disabled` and `disabled-text` I can do that in this PR or if we want to discuss a little more I can make another issue

ericsoderberghp: Let's handle that as a separate issue, to ensure we add it consistently to grommet.

ericsoderberghp (approved):

ericsoderberghp:
Before merging this, I'd like to have a conversation about how we communicate this change, since it could be perceived as a breaking change by consumers of the `hpe` theme.

halocline:
> Before merging this, I'd like to have a conversation about how we communicate this change, since it could be perceived as a breaking change by consumers of the `hpe` theme.

Just made a [similar comment here](https://github.com/grommet/hpe-design-system/pull/2349/files#r805938859) when reviewing the What's New work.

britt6612:
> similar comment here

thank you Ill add more details in there.. I'm wondering if having it in what's new is good enough for all of our users?

ericsoderberghp:
> > similar comment here
> 
> thank you Ill add more details in there.. I'm wondering if having it in what's new is good enough for all of our users?

I think we should also explicitly mention it in office hours.

PR #237, opened 2022-03-07
Change Select options font weight

britt6612:
#### What does this PR do?
This PR changes the `text` in the select options to have a weight of 500
site
match what is on figma 
closes https://github.com/grommet/hpe-design-system/issues/1242

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -201,7 +201,7 @@ export const hpe = deepFreeze({
         radius: '0px',
       },
       font: {
-        weight: 400,
+        weight: 500,
       },
     },
     active: {
@@ -950,11 +950,7 @@ export const hpe = deepFreeze({
       down: FormDown,
       up: FormUp,
     },
-    options: {
-      text: {
-        weight: 500,
-      },
-    },
+    options: undefined,
   },
   spinner: {
     container: {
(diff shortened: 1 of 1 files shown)


britt6612:
> Do you have a screenshot from site test?
Screen Shot 2022-03-07 at 1 06 23 PM

src/js/themes/hpe.js
```diff
@@ -950,7 +950,11 @@ export const hpe = deepFreeze({
       down: FormDown,
       up: FormUp,
     },
-    options: undefined,
+    options: {
```
taysea, line 953: I think we want to leave this as undefined because we have "option" button kind which is used for this kind of button and applies some other styling as well. Thinking we should adjust the label weight in that object instead.

taysea: I believe [this line](https://github.com/grommet/grommet-theme-hpe/pull/237/files#diff-0ea9b1217744b978fa7c1ac3ab44c113973e55fbe073fc22011857993680bb57R204) should change to 500.

britt6612: good insight I was not aware we had that button kind I will update.

taysea (approved):
LGTM.

ericsoderberghp (approved):

PR #241, opened 2022-03-11
add theme for banner notification

britt6612:
#### What does this PR do?
Adds theme for banner notification
DS site
view changes https://deploy-preview-2363--keen-mayer-a86c8b.netlify.app/templates/global-banner-notifications?q=glo
completes banner notification issue
backward compatible
yes

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -846,43 +846,23 @@ export const hpe = deepFreeze({
         round: 'none',
       },
     },
-    critical: {
-      toast: {
-        background: 'background-front',
-      },
-    },
     message: {
       color: { dark: 'text-strong' },
     },
     warning: {
       background: 'validation-warning',
-      toast: {
-        background: 'background-front',
-      },
     },
     normal: {
       background: 'validation-ok',
-      toast: {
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -833,6 +833,59 @@ export const hpe = deepFreeze({
...
+      round: 'xsmall',
+    },
+    global: {
+      container: {
+        round: 'none',
+      },
+    },
+    critical: {
+      background: 'validation-critical',
+      toast: {
```
taysea, line 851: I think in latest grommet stable we now have a defined color for notification[status]global.background -- which we might need to override to ensure out validation colors are coming through.

Can you double check that?

britt6612: yes Ill check now

britt6612: yes the critical is coming through so we don't need that just the rest

britt6612: looks like we dont need `toast` background colors either

jcfilben (approved):
Looks good!

Side note: I think we need to updated the grommet-site docs to include `notification.global`

PR #242, opened 2022-03-16
add background for toast

britt6612:
#### What does this PR do?
fix color for Toast notifications 
design system site
brought up in slack
change backward
what it was before

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -851,18 +851,33 @@ export const hpe = deepFreeze({
     },
     warning: {
       background: 'validation-warning',
+      toast: {
+        background: 'background-front',
+      },
     },
     normal: {
       background: 'validation-ok',
+      toast: {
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -851,18 +851,33 @@ export const hpe = deepFreeze({
     },
     warning: {
       background: 'validation-warning',
```
taysea, line 853: we also need to add:
```
global: {
   background: 'validation-warning',
}
```
and the equivalent for all the statuses. right now it's pulling the default from base.js which is a different color than our validation colors.

britt6612: thank you for clarifying I pushed up changes

taysea (approved):
LGTM. Thanks for fixing!

jcfilben (approved):

PR #261, opened 2022-08-02
First cut at Button changes for the HPE Brand refresh

MikeKingdom:
A first cut changes to the theme to add a CTA kind for Button and round the buttons.

#### What does this PR do?
- Removes the HPE specific Button border radius settings and lets the base rounded buttons be the style.
- Adds  a `cta` kind for the Call To Action button style with the brand coloring and a hover style similar to the primary button. Note that primary buttons and secondary button borders are a brighter green than the brand color.

What this doesn't address:
- Embedding the HPE element icon within the styling
- Some form of animation of the element on hover

https://github.com/grommet/hpe-design-system/issues/2779
https://github.com/grommet/hpe-design-system/issues/2774

image

image
image
image
image

image

image

Not really a breaking change but it is a significant visual difference to existing buttons including drops' hovers and such.

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -15,12 +15,12 @@ import { colors } from './colors';
 
 const baseSpacing = 24;
 
-const isObject = (item) =>
+const isObject = item =>
   item && typeof item === 'object' && !Array.isArray(item);
 
-const deepFreeze = (obj) => {
+const deepFreeze = obj => {
   Object.keys(obj).forEach(
-    (key) => key && isObject(obj[key]) && Object.freeze(obj[key]),
+    key => key && isObject(obj[key]) && Object.freeze(obj[key]),
   );
   return Object.freeze(obj);
 };
@@ -202,6 +202,16 @@ export const hpe = deepFreeze({
     },
   },
   button: {
+    cta: {
+      background: {
+        color: 'brand',
+      },
+      border: undefined,
+      color: 'white', // 'text-strong',
+      font: {
+        weight: 700,
+      },
+    },
     default: {
       color: 'text-strong',
       border: undefined,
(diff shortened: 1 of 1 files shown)


ericsoderberghp:
Nice progress. A few initial comments:
- I think the icon-only buttons should not have the extra horizontal padding, especially as shown in the DataTable header.
- The buttons with both a label and an icon look too cramped vertically.
- I'm wondering if we should special case the height only for the `cta` kind, leave the height as it exists today for all other kinds, but add the rounding across the board

Also, can we set this up via CircleCi to automatically publish a "brand-refresh-stable" branch that would allow use via `https://github.com/grommet/grommet-hpe-theme/tarball/brand-refresh-stable`? That would allow us to wire up the design system site and grommet designer to use it so we could see it in action more broadly.

taysea (approved):
Looks good. I propose we merge into NEXT so we can play around with it more on the site.

src/js/themes/hpe.js
```diff
@@ -202,6 +207,33 @@ export const hpe = deepFreeze({
...
+    cta: {
+      background: {
+        color: 'brand',
+      },
+      border: undefined,
+      color: 'white', // 'text-strong',
+      font: {
+        weight: 700,
+      },
+      extend: props => {
```
halocline, line 219: Comment for clarity.

```suggestion
     // TODO: Enhance Grommet to allow theme to set `button.{kind}.icon` which is an SVG.
     // With this enhancement, the `extend` property definition should be able to be removed.
      extend: props => {
```

halocline (commented):
@MikeKingdom - I'm missing where the border-radii are being updated. Can you point me to where they are being set?

MikeKingdom:
> @MikeKingdom - I'm missing where the border-radii are being updated. Can you point me to where they are being set?

@halocline  base.js in grommet. It's built into the 'size' part of button. hpe.js was basically overriding it.

halocline:
> > @MikeKingdom - I'm missing where the border-radii are being updated. Can you point me to where they are being set?
> 
> @halocline base.js in grommet. It's built into the 'size' part of button. hpe.js was basically overriding it.

Ahhh, got it. Thanks!

halocline (approved):
Let's merge and continue refining from here.

FYI, I've added a couple deliverables to [this issue](https://github.com/grommet/hpe-design-system/issues/2774) which ask for "Grommet enhancement issues" to be created once we explore implementing via theme alone.

PR #270, opened 2022-08-18
Makes background gradients and textures available to theme

halocline:
#### What does this PR do?

Closes https://github.com/grommet/hpe-design-system/issues/2824 by adding a `backgrounds` theme object with namespaces.

### DIFF
--- a/src/js/themes/backgrounds.js
+++ b/src/js/themes/backgrounds.js
@@ -1,0 +1,64 @@
+export const backgrounds = {
+  gradients: {
+    'orange-yellow': `linear-gradient(
+      45deg,
+      hsl(22deg 100% 50%) 0%,
+      hsl(34deg 100% 50%) 50%,
+      hsl(46deg 100% 50%) 100%
+    );`,
+    'purple-blue': `linear-gradient(
+      45deg,
+      hsl(281deg 100% 63%) 1%,
+      hsl(227deg 83% 58%) 50%,
+      hsl(174deg 69% 53%) 99%
+    );`,
+    'purple-blue-yellow': `linear-gradient(
+      225deg,
+      hsl(263deg 82% 55%) 0%,
+      hsl(196deg 72% 53%) 25%,
+      hsl(171deg 80% 63%) 50%,
+      hsl(138deg 93% 68%) 75%,
+      hsl(47deg 99% 50%) 100%
+    );`,
+    'purple-magenta-yellow': `linear-gradient(
+      45deg,
+      hsl(274deg 100% 50%) 0%,
+      hsl(340deg 100% 50%) 50%,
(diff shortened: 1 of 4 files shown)


taysea (approved):

ericsoderberghp:
I was hoping that these new backgrounds could be used via something like `<Box background="datawave-white-2" />`. But, it looks like this is creating some separate imports.
Also, would it make any sense for any of these to leverage the `{dark: ..., light: ...}` construction to work well with automatic theme mode changes?

taysea:
> I was hoping that these new backgrounds could be used via something like `<Box background="datawave-white-2" />`. But, it looks like this is creating some separate imports. Also, would it make any sense for any of these to leverage the `{dark: ..., light: ...}` construction to work well with automatic theme mode changes?

@halocline can provide more detail, this is the approach we were first aiming to do. There were a couple items re: image positioning that were missing when passing it like this. We went for a first approach of what could be implemented given functionality now, but I think we could (and maybe should) definitely enhance Box to make it work like this.

halocline:
> Also, would it make any sense for any of these to leverage the `{dark: ..., light: ...}` construction to work well with automatic theme mode changes?

We can make that adjustment.

halocline:
> > I was hoping that these new backgrounds could be used via something like `<Box background="datawave-white-2" />`. But, it looks like this is creating some separate imports. Also, would it make any sense for any of these to leverage the `{dark: ..., light: ...}` construction to work well with automatic theme mode changes?
> 
> @halocline can provide more detail, this is the approach we were first aiming to do. There were a couple items re: image positioning that were missing when passing it like this. We went for a first approach of what could be implemented given functionality now, but I think we could (and maybe should) definitely enhance Box to make it work like this.

I agree that `<Box background="datawave-white-2" />` is what we want.

I need to investigate a bit more, but I believe we need to make a couple tweaks to `background.js` to accommodate. There were cases where `background-size` was not being set to `cover` as desired.

Additionally, I wanted to propose supporting `background-clip` to enable some of the typography treatments. Something like:
```
<Box
  background={{
      image: 'datawave-multi-1',
      clip: true,
   }}
>
   <Heading margin="none">Modernize</Heading>
       <Text margin="none" size="xlarge">
           edge to cloud
       </Text>
</Box>
```
and and potentially the ability to rotate the gradients:
```
<Box
     background={{
         image: 'purple-blue-yellow',
         clip: true,
         rotate: 90,
     }}
>
    <Heading margin="none">Modernize</Heading>
        <Text size="xlarge" margin="none">
            edge to cloud
        </Text>
</Box>
```


Lastly, this proposes a new theme property `theme.global.backgrounds`. Are you okay with that direction?

ericsoderberghp:
I'm very supportive of making the necessary changes in grommet to accommodate these things.

PR #288, opened 2022-10-28
Use Brand Central font versions

taysea:
#### What does this PR do?

Based on outcomes of https://github.com/grommet/hpe-design-system/issues/2990, reverting the theme to point to brand central font version. The URLs were provided by Artur's team. I had to explicitly define regular as font-weight 400 otherwise it wasn't applying it.

Tested locally in the site on Mac Chrome, Windows Edge:

Mac Chrome:
Screen Shot 2023-01-27 at 3 36 21 PM

New font, Windows Edge (notice how the "i" is differentiable from "l", j has a visibly separate dot)

Screen Shot 2023-01-27 at 3 38 54 PM

Closes https://github.com/grommet/hpe-design-system/issues/2990

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -106,28 +106,37 @@ export const hpe = deepFreeze({
       face: `
         @font-face {
           font-family: "Metric";
-          src: url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe/MetricHPE-Web-Regular.woff") format('woff');
+          src: url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe-web/MetricHPE-Web-Regular.woff2") format('woff2'),
+               url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe-web/MetricHPE-Web-Regular.woff") format('woff');
         }
         @font-face {
           font-family: "Metric";
-          src: url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe/MetricHPE-Web-Bold.woff") format('woff');
+          src: url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe-web/MetricHPE-Web-Regular.woff2") format('woff2'),
+               url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe-web/MetricHPE-Web-Regular.woff") format('woff');
+          font-weight: 400;
+        }
+        @font-face {
+          font-family: "Metric";
+          src: url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe-web/MetricHPE-Web-Bold.woff2") format('woff2'),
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -106,31 +106,26 @@ export const hpe = deepFreeze({
       face: `
         @font-face {
           font-family: "Metric";
-          src: url("https://d3hq6blov2iije.cloudfront.net/fonts/HPEXS-Metric-Fonts/MetricHPEXS-Regular.woff2") format('woff2'),
-               url("https://d3hq6blov2iije.cloudfront.net/fonts/HPEXS-Metric-Fonts/MetricHPEXS-Regular.woff") format('woff');
+          src: url("https://hpefonts.s3.amazonaws.com/web/MetricHPE-Web-Regular.woff") format('woff');
```
halocline, line 109: Hmmm.... I'm wondering if we can connect with someone who owns this S3 bucket? Since this is not on a CDN, folks accessing any of our apps from Ireland, Germany, Japan, India, Australia will be having to make intercontinental network requests.

hpe.com is making use of a CDN at this path: https://www.hpe.com/apps/hpeweb-ui/fonts/Metric-Regular.woff2, but notice it is not labelled MetricHPE.

Perhaps the Brand folks can help get us connected.

taysea: Ooo thanks for catching that. I hadn't realized. I would agree we likely want to use the CDN you shared above

taysea:
Interestingly it seems like that CDN has Metric-Regular, -Light, -Semibold but is missing -Medium and -Bold.

halocline:
> Interestingly it seems like that CDN has Metric-Regular, -Light, -Semibold but is missing -Medium and -Bold.

I think we should at least try to have a conversation with the team managing. Let's do that and evaluate from there.

src/js/themes/hpe.js
```diff
@@ -106,31 +106,28 @@ export const hpe = deepFreeze({
       face: `
         @font-face {
           font-family: "Metric";
-          src: url("https://d3hq6blov2iije.cloudfront.net/fonts/HPEXS-Metric-Fonts/MetricHPEXS-Regular.woff2") format('woff2'),
-               url("https://d3hq6blov2iije.cloudfront.net/fonts/HPEXS-Metric-Fonts/MetricHPEXS-Regular.woff") format('woff');
+          src: url("https://www.hpe.com/h41225/hfws-static/fonts/metric-hpe/MetricHPE-Web-Regular.woff") format('woff');
```
halocline, line 109: We should be providing the woff2 versions for all of these fonts. As far as I can tell, we have them all in the email, but the file naming convention is inconsistent.

taysea: From the urls, I interpreted that MetricHPE-Web was the path we wanted to use, but just sent an email to Artur to follow up about any differences between MetricHPE and Metric from the urls.

ericsoderberghp (approved):

PR #298, opened 2022-12-14
add border weak to spinner

britt6612:
#### What does this PR do?
changes spinner to be `border-weak` opposed too `background-contrast`
aries site
match figma designs
tested in site

https://user-images.githubusercontent.com/42451602/207487766-c004012e-713e-4b2e-8627-35799e302266.mov

yes change to spinner


taysea:
Maybe this is just my opinion, but I prefer having a more subtle "background-contrast" color for the remaining part of the border because it allows more focus on the green part/conveys more motion. Soft opinion and no need to hold things up, but just a reaction

vavalos5:
> Maybe this is just my opinion, but I prefer having a more subtle "background-contrast" color for the remaining part of the border because it allows more focus on the green part/conveys more motion. Soft opinion and no need to hold things up, but just a reaction

Thanks for the honest feedback Taylor! When Brittany and I were taking a look at this in Figma, the `background-contras`t color was not visible within the figma file. Our Figma file by default had the `border- weak` color instead, making it look more visible. It had been interesting seeing that it looks different in code. We had decided to move forward with the `border-weak` color to ensure it was more visible to users, especially those with color impairments.

taysea (approved):

PR #305, opened 2023-01-28
update with latest on-brand specs

halocline:
#### What does this PR do?

Implements latest brand direction for Product and Web theme experiences.

https://github.com/grommet/hpe-design-system/issues/3102

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -27,6 +27,30 @@ const deepFreeze = (obj) => {
   return Object.freeze(obj);
 };
 
+const primaryBackground = (props) => {
+  let style = '';
+  if (!props.active) {
+    style += !props.colorValue
+      ? `background:
+linear-gradient(70deg, transparent,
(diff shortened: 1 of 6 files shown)


src/js/themes/hpe.js
```diff
@@ -338,50 +378,18 @@ export const hpe = deepFreeze({
       },
     },
     disabled: {
```
taysea, line 380: I actually think we can remove this entire block of "disabled". `opacity: 0.3` comes from grommet theme (we can leave here to be explicit if we want) and the `undefined` overrides for button kinds/background/etc. were only there to override what we had in grommet-theme-hpe (which will now be removed).

taysea (approved):
One small comment about `button.disabled` cleanups but it doesn't have to block merging this PR. we can cleanup later.

src/js/themes/hpe.js
halocline: done.

taysea (approved):
LGTM. Thanks for transferring it over here.

PR #324, opened 2023-03-01
initial audit theme and rules

halocline:
#### What does this PR do?

Creates an audit theme for use when conducting brand audits. Callers will reference the AUDIT-stable branch from their `package.json` and then apply the audit test by using the `audit` theme.
```
import { audit } from 'grommet-theme-hpe';

...

<Grommet theme={audit} ... />

```

https://github.com/grommet/hpe-design-system/issues/3154

Backwards compatible.


taysea (approved):
This looks like a good structural set up.

PR #329, opened 2023-03-14
Set default radius to 2em

taysea:
#### What does this PR do?

Instead of defining on each kind, set default button radius to 2em. This allows us to leave `default` button radius as `undefined` so the badge aligns to the label/icon instead.

Screen Shot 2023-03-13 at 6 48 48 PM

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -322,6 +322,7 @@ export const hpe = deepFreeze({
     },
     default: {
       color: 'text-strong',
+      // necessary so badge overlaps label/icon
       border: undefined,
       font: {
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -458,7 +446,7 @@ export const hpe = deepFreeze({
     size: {
       small: {
         border: {
-          radius: '6px',
+          radius: '2em',
```
ericsoderberghp, line 449: Just curious, could we set it at `theme.button.border.radius`?

taysea: I'll see if we can do this

PR #338, opened 2023-03-22
Update DataTable contract/expand icons

taysea:
#### What does this PR do?

Updates the DataTable contract/expand icons for `groupBy` to use Down/Up instead of form icons. Height "medium" ensures the icon is vertically-aligned with the other content in the row.

Screen Shot 2023-03-21 at 6 08 57 PM


halocline (commented):
While we are here...

halocline (approved):

PR #342, opened 2023-03-23
Aligned webpack externals with peer dependencies

ericsoderberghp:
#### What does this PR do?

Aligned webpack externals with peer dependencies

grommet-designer dynamic theme loading

backwards compatible

I think this should be automatically handled by the final app bundler packaging. But, would be good to double check with something like the hpe-design-system site.

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -1,17 +1,19 @@
 // (C) Copyright 2020-2021 Hewlett Packard Enterprise Development LP
 import React from 'react';
 import { css } from 'styled-components';
-import { Ascending } from 'grommet-icons/icons/Ascending';
-import { Blank } from 'grommet-icons/icons/Blank';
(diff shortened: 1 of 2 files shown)


src/js/themes/hpe.js
```diff
@@ -851,7 +851,7 @@ export const hpe = deepFreeze({
       container: {
         gap: 'xsmall',
       },
-      icon: <CircleAlert size="small" />,
+      icon: () => <CircleAlert size="small" />,
```
taysea, line 854: @ericsoderberghp Are these changes also necessary? Just asking because no issues have been flagged by them prior

ericsoderberghp: I'm just aligning all of the icons to work the same way. I don't know that it really matters. But, it seemed like this might be slightly lighter weight.

taysea (approved):
Thanks for catching this.

PR #343, opened 2023-03-23
Fix small icon-only padding and icon rendering

taysea:
#### What does this PR do?

Adjusts small icon-only padding to `10px` to ensure `36px` dimensions. Reverts changes to icon format from #342 which was causing error/not rendering icon.

Screen Shot 2023-03-23 at 9 18 33 AM
Screen Shot 2023-03-23 at 9 18 12 AM

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -313,7 +313,7 @@ export const hpe = deepFreeze({
       background: { color: 'brand' },
       color: 'text-primary-button',
       font: { weight: 'bold' },
-      icon: Hpe,
+      icon: <Hpe />,
       reverse: true,
       extend: (props) => primaryBackground(props),
     },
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -313,7 +313,7 @@ export const hpe = deepFreeze({
       background: { color: 'brand' },
       color: 'text-primary-button',
       font: { weight: 'bold' },
-      icon: Hpe,
+      icon: <Hpe />,
```
halocline, line 316: Need to merge master? Seems like these revert what Eric just submitted?

taysea: Reverts were intended due to errors when rendering icons. The fix to add `react` as an external in package.json wasn't reverted though, since that fix is necessary.

halocline (approved):

PR #344, opened 2023-03-27
AUDIT branch - Export audit as hpe

halocline:
#### What does this PR do?

Export audit as hpe so that caller doesn't have to do any other config changes.


taysea (approved):
Nice cleanup. Makes sense.

PR #347, opened 2023-04-07
fix template literal interpolations

halocline:
#### What does this PR do?

When upgrading to 5.0.1, GLCP reported an error in their Jest test suite: `undefined:1:1232: property missing ':'` related to a checkbox test.

https://github.com/styled-components/jest-styled-components/issues/97#issuecomment-364169127 pointed to https://styled-components.com/docs/api#css as the solution.

Most discussion points to the styles being rendered correctly in browsers, but throwing errors in certain Jest environments.

Built project locally and consumed in GLCP/frontend/ui Jest test suite.

Reported by Annie.


taysea (approved):
Thanks for the fix and references to jest discussions.

PR #350, opened 2023-04-11
Changed release-gh-pages to use plain name

ericsoderberghp:
#### What does this PR do?

Changed release-gh-pages to use plain name

waiting for CircleCI

The GitHub pages are used by grommet-designer and the "hpe" theme should be associated with the latest default path.

coordinated with grommet-designer

shouldn't be needed


taysea (commented):
Confirming the intent is to remove the "2" and also remove the JSON files?

taysea (approved):

PR #384, opened 2024-03-01
Add use client to index.js

taysea:
#### What does this PR do?

Add use client to index.js to support use in NextJS App router which assumed SSR unless specified.


halocline (approved):
🙏

PR #397, opened 2024-05-23
Styled components 6 support

jcfilben:
#### What does this PR do?
Adds `&` in front of `:hover`. In styled components 5 this was implicitly included if it wasn't already there but in styled components 6 this needs to be manually added

backwards compatible

styled components v6 support


britt6612 (approved):
Nice catch!

PR #426, opened 2025-02-06
Fix states for select.option

taysea:
#### What does this PR do?

"option" styles are handled as separate logic to buttonKindTheme and buttonStates theme. This PR:
- Removes unused block of code in those functions that checked for "if kind === 'option'" even though the kind array we're mapping through doesn't use "option"
- Adds appropriate missing states for option styles: disabled, active (which applies when "option" has keyboard focus)
- Fixes incorrect reference to textColor token

Locally in sandbox/grommet-app

BEFORE

https://github.com/user-attachments/assets/2971127b-671a-4587-87ca-14b4e24aa787

AFTER

https://github.com/user-attachments/assets/8565a6d4-1fe8-451a-9323-cb7ec5da5f3d

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -799,7 +799,6 @@ const buildTheme = (tokens, flags) => {
         extend: ({ kind, theme }) =>
           kind === 'option' &&
           `
-          background: transparent;
           &[aria-selected="true"] { background: ${getThemeColor(
             components.hpe.select.default.option.selected.rest.background,
             theme,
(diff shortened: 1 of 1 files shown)


jcfilben:
Looks like we might want to make a change for SelectMultiple as well. This is what it looks like when an option is selected and disabled
Screenshot 2025-02-06 at 10 10 26 AM

jcfilben:
Select changes look good though

MikeKingdom (approved):
LGTM

taysea:
> Looks like we might want to make a change for SelectMultiple as well. This is what it looks like when an option (in this case the third option) is selected and disabled Screenshot 2025-02-06 at 10 10 26 AM

This is fixed by https://github.com/grommet/grommet-theme-hpe/pull/428

jcfilben (approved):

PR #432, opened 2025-02-06
Fix primitive color reference

taysea:
#### What does this PR do?

In hpe-design-tokens we're not representing these "!" colors because they may not carry through to next brand iteration, so we did not want to carry over any baggage. We then removed any unused primitive colors from hpe-design-tokens, these two primitive colors were not used elsewhere in the semantic set so they were removed. For backwards compatibility, hard coding the existing value.

All other "!" primitive colors still exist.

Also fixing one other missing color reference. We removed "text-xweak" from design tokens for accessibility, but since we never marked it as "deprecated" in any previous grommet-theme-hpe, we will point it to "text-weak" so there is a value. In future major version of tokens we'll remove.

### DIFF
--- a/src/js/themes/colors.js
+++ b/src/js/themes/colors.js
@@ -79,6 +79,10 @@ export const colors = {
     dark: dark.hpe.color.text.default,
     light: light.hpe.color.text.default,
   },
+  'text-xweak': {
+    dark: dark.hpe.color.text.weak,
+    light: light.hpe.color.text.weak,
+  },
   border: {
     dark: dark.hpe.color.border.default,
     light: light.hpe.color.border.default,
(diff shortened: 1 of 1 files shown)


src/js/themes/colors.js
```diff
@@ -79,6 +79,10 @@ export const colors = {
     dark: dark.hpe.color.text.default,
     light: light.hpe.color.text.default,
   },
+  'text-xweak': {
```
jcfilben, line 82: Should we add a comment to remind ourselves to deprecate this?

taysea: Done.

britt6612 (approved):

PR #434, opened 2025-02-06
Fix clear-selection focus style

taysea:
#### What does this PR do?

Fixes clear selection :focus styles to accommodate/align to the "padding" around options.

Locally in sandbox/grommet-app

https://github.com/user-attachments/assets/13e73282-8dea-43e4-9d3c-ff4547378a95

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -2430,9 +2430,6 @@ const buildTheme = (tokens, flags) => {
             horizontal: '12px',
             vertical: '6px',
           },
-          margin: {
-            horizontal: components.hpe.select.default.medium.drop.paddingX,
-          },
           hover: {
             background: 'background-hover',
           },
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -2443,6 +2440,29 @@ const buildTheme = (tokens, flags) => {
           weight: components.hpe.button.default.rest.fontWeight,
         },
       },
+      container: {
+        // Applying spacing on Select "Clear selection" button, then placing focus styles on the inner container div
+        extend: ({ theme }) =>
+          `
+          button[aria-label*="Or, press"] {
```
MikeKingdom, line 2447: I wouldn't think this would work with localization (can't this message be changed via `messages`)?

britt6612: Im not sure if you can change via `messages` but       
```
<StyledButton
        a11yTitle={`${buttonLabel}. Or, press ${
          position === 'bottom' ? 'shift tab' : 'down arrow'
        } to move to select options`}

```

this `button` does have `...rest` on it so if someone passes a different a11yTitle Im assuming it would override whats there

MikeKingdom: Yeah and the fact it has hard coded text is really a defect since it isn't localizable. In general it's never a good idea to compare against a text label. Too fragile and easy to change in the component and not realize this is here.

taysea: Totally agree here/was something Jessica and I chatted through initially that this isn't a good long term solution, but working with what is available at the moment.

This button doesn't currently support internationalization via grommet messages (@MikeKingdom correct me if I'm wrong there) and the "...rest" is just local to the file so no caller could override the a11yTitle.

I plan to file a grommet issue to enhance the theme object to give us control where we wouldn't need to style this via `extend`, but given the string is hard coded this feels like the best selector to use vs trying to do something more specific to DOM structure.

Let me know if this is reasonable for the time being.

MikeKingdom: Probably fine for now. We need to make sure we don't forget about it or let it linger. We do have partners that localize and these kinds of things do come back around.

jcfilben: ^ agree

jcfilben (approved):

PR #454, opened 2025-03-20
Ensure stable branches have package.json

taysea:
#### What does this PR do?

At some point in time, our stable branches stopped having a package.json which means that repositories (like design system site) trying to test the stable branch can't use it.

This PR ensures that stable branches always copy over the package.json. Additionally, updates the stable branches to mirror the production structure where theme files are nested under "dist" folder.

Simplifies file structure by removing redundant scripts and instead enhancing the standard release-stable script to accept a branch.

Locally, ran each release stable script and confirmed the correct branch was targeted.


jcfilben (approved):

MikeKingdom (approved):
Looks good!

PR #467, opened 2025-05-07
Update Tab styles

taysea:
#### What does this PR do?

Updates Tab styles in response to theme feedback.

Designs can be found: https://www.figma.com/design/ine5KzQ6rA41GboLZuG05A/Theme-feedback?node-id=454-34121&t=uvwx56V8XTiHzt7L-4

Locally in DS site: https://github.com/grommet/hpe-design-system/pull/5129

https://github.com/user-attachments/assets/ada3f2e0-4327-4828-b47a-b13852279215

Closes https://github.com/grommet/hpe-design-system/issues/5130

### DIFF
--- a/.changeset/four-mirrors-yell.md
+++ b/.changeset/four-mirrors-yell.md
@@ -1,0 +1,5 @@
+---
+"grommet-theme-hpe": minor
+---
+
+- Updated Tab styles to use underline marker instead of background fill to indicate the selected tab. This removes potential conflict in visual hierarchy with primary buttons.
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -2270,18 +2270,16 @@ const buildTheme = (tokens, flags) => {
     tab: {
       color: 'text',
       active: {
-        background: 'background-selected-primary-strong',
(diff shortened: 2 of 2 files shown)


britt6612:
I see now when we hove the text changes from regular font to `bold` is that enough for `wcag 1.4.11: Non-text Contrast` 

https://github.com/user-attachments/assets/240e017a-1055-48b2-8d1c-c7ab6b3acc46

just one small question otherwise looks good

jcfilben:
> I see now when we hove the text changes from regular font to bold is that enough for wcag 1.4.11: Non-text Contrast

I would expect the cursor to change when we hover over the tab to indicate that it is clickable and for us to use that as an indicator rather than the hover styling

britt6612:
> > I see now when we hove the text changes from regular font to bold is that enough for wcag 1.4.11: Non-text Contrast
> 
> I would expect the cursor to change when we hover over the tab to indicate that it is clickable and for us to use that as an indicator rather than the hover styling

yes I didnt know if that change was intentional I believe the curser changes now prior to these changes

halocline (approved):

PR #479, opened 2025-05-29
Add support for React 19 to peerDeps

taysea:


MikeKingdom (approved):
Thanks! I'll try it out after it merges and ends up in stable

Issue #461, opened 2025-04-24
Range Input - WCAG Rule 1.4.11 Non-text Contrast


KennyAtHPE:
- Filled portion of the track vs non filled portion fails contrast ratio requirements when range input is placed on top of background default, front, and back.
- Filled portion of the track fails contrast ratio requirements when range input is placed on top of background back.
- (Dark mode) Thumb vs non filled portion of the track fails contrast ratio requirements when range input is placed on top of background default, back, and front
- (Dark mode) Focus indicator vs non filled portion of the track fails contrast ratio when range input is placed on top of background default, back, and front
- (Dark mode) Focus indicator fails contrast ratio when range input is placed on top of background default, back, and front

jcfilben:
Confirm that the latest token updates have resolved this issue

jcfilben:
I checked with the latest token colors and here were the results:

### Light mode

#### B=background-front & background
Pass - green vs grey
Pass - green vs background
Fail - grey vs background

#### background-back
Pass - green vs grey
Pass - green vs background
Fail - grey vs background

### Dark Mode

#### background-front
Fail - green vs grey
Pass - green vs background
Fail - grey vs background

#### background & background-back
Fail - green vs grey
Pass - green vs background
Fail - grey vs background

taysea:
Reopening until we release in grommet-theme-hpe and update docs.

PR #482, opened 2025-06-10
Fix RangeInput styles to meet WCAG AA requirements

taysea:
#### What does this PR do?

Adds "dot" treatment on unselected end of track to ensure appropriate affordance for end of track. At 3px by 3px, a minimum contrast ratio of 3:1 is required, which we more than exceed with the chosen semantic color.

rest + disabled states

Screenshot 2025-06-10 at 12 39 31 PM

Tested contrast ratios.

Screenshot 2025-06-10 at 12 38 49 PM

Closes https://github.com/grommet/grommet-theme-hpe/issues/461

Backwards compatible

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -2145,7 +2145,7 @@ const buildTheme = (tokens, flags) => {
           content: '';
           width: 3px;
           height: 3px;
-          border-radius: 9999px;
+          border-radius: ${large.hpe.radius.full};
           right: 0;
           top: 50%;
           transform: translateY(-50%);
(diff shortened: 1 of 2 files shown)


src/js/themes/hpe.js
```diff
@@ -2138,6 +2138,23 @@ const buildTheme = (tokens, flags) => {
           color: { light: 'rgb(245, 245, 245)', dark: 'rgb(44, 44, 44)' },
         },
       },
+      extend: ({ disabled, theme }) => `
+        &::before {
+          display: block;
+          position: absolute;
+          content: '';
+          width: 3px;
```
britt6612, line 2146: is there not a token for `3px` ? or some size we can use here?

taysea: Done. Using a primitive value from our scale since there are not semantic tokens that would be appropriate.

halocline (approved):

britt6612 (approved):

PR #485, opened 2025-06-18
Range selector edge size

britt6612:
#### What does this PR do?
This PR uses the latest changes in grommet https://github.com/grommet/grommet/pull/7676
to make the edge sizes large to meet WCAG requirements. Now the edge size will be 24X24
In DS site 
RangeSelector fails 2.5.5

The circular buttons on either side of the slider do not meet the minimum interactive target size of 24×24.

Currently we have new designs for RangeSelector but in order to meet the WCAG we are okay with this change for the mean time. 
Closes https://github.com/grommet/grommet/issues/7669
breaking

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -8,7 +8,7 @@ import {
   small as localSmall,
   global as localGlobal,
   components as localComponents,
-  primitives as localPrimitives
+  primitives as localPrimitives,
 } from 'hpe-design-tokens/grommet';
 import { Ascending } from 'grommet-icons/icons/Ascending';
 import { Blank } from 'grommet-icons/icons/Blank';


src/js/themes/hpe.js
```diff
@@ -2139,6 +2139,11 @@ const buildTheme = (tokens, flags) => {
         },
       },
     },
+    rangeSelector: {
+      edge: {
+        size: `${primitives.hpe.base.dimension[600]}`, // 24px
```
taysea, line 2144:
```suggestion
        size: ${primitives.hpe.base.dimension[600]}, // 24px
```
We should be able to do this I think

britt6612: this created another issue so I put back

PR #488, opened 2025-07-10
Add `global.focus.inset`

taysea:
#### What does this PR do?

Adds `global.focus.inset` for cases like "expand" control in DataTable where focus needs to be inset in the container to show.

**Note to reviewers**

@jcfilben @britt6612 We can discuss if this should be a "major" change (since it won't actually come through unless someone is on the grommet version that includes this enhancement) or a "minor" change (it would come through for teams that are on the latest, but it wouldn't break anything for teams on older versions).

Locally in Design System site.

Screenshot 2025-07-10 at 4 10 19 PM
Screenshot 2025-07-10 at 4 10 10 PM

Closes https://github.com/grommet/grommet/issues/7336

### DIFF
--- a/.changeset/heavy-fireants-walk.md
+++ b/.changeset/heavy-fireants-walk.md
@@ -2,4 +2,4 @@
 "grommet-theme-hpe": minor
 ---
 
-- Added `global.focus.inset` for cases like DataTable expand control which leverages `focusIndicator="inset"` internally. Supported when using grommet >=2.X.0.
+- Added `global.focus.inset` for cases like DataTable expand control which leverages `focusIndicator="inset"` internally. Supported when using grommet >=2.48.0.
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -630,21 +630,19 @@ const buildTheme = (tokens, flags) => {
         },
         twoColor: true,
         inset: {
-          inset: {
-            border: undefined,
-            outline: {
-              color: global.hpe.focusIndicator.outline.color,
-              size: global.hpe.focusIndicator.outline.width,
-              offset: `-${global.hpe.focusIndicator.outline.width}`,
-            },
-            shadow: {
-              color: focusBoxShadowParts[focusBoxShadowParts.length - 1],
-              size: '4px',
-              blur: '0px',
-              inset: true,
-            },
-            twoColor: true,
+          border: undefined,
+          outline: {
+            color: global.hpe.focusIndicator.outline.color,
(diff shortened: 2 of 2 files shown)


taysea:
@jcfilben @britt6612 We can discuss if this should be a "major" change (since it won't actually come through unless someone is on the grommet version that includes this enhancement) or a "minor" change (it would come through for teams that are on the latest, but it wouldn't break anything for teams on older versions).

src/js/themes/hpe.js
```diff
@@ -629,6 +629,23 @@ const buildTheme = (tokens, flags) => {
           size: focusBoxShadowParts[focusBoxShadowParts.length - 2],
         },
         twoColor: true,
+        inset: {
+          inset: {
```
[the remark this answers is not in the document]
taysea: Fixed!

jcfilben (commented):
Looks good!

> We can discuss if this should be a "major" change (since it won't actually come through unless someone is on the grommet version that includes this enhancement) or a "minor" change (it would come through for teams that are on the latest, but it wouldn't break anything for teams on older versions).

I would be okay with us considering this as a minor change since we are not requiring teams to do a grommet version upgrade

jcfilben (approved):

britt6612:
> Looks good!
> 
> > We can discuss if this should be a "major" change (since it won't actually come through unless someone is on the grommet version that includes this enhancement) or a "minor" change (it would come through for teams that are on the latest, but it wouldn't break anything for teams on older versions).
> 
> I would be okay with us considering this as a minor change since we are not requiring teams to do a grommet version upgrade

plus 1 here I think a minor change is fine for this is "fixing" a bug and not a new enhancement

.changeset/heavy-fireants-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"grommet-theme-hpe": minor
+---
+
+- Added `global.focus.inset` for cases like DataTable expand control which leverages `focusIndicator="inset"` internally. Supported when using grommet >=2.X.0.
```
britt6612, line 5: do we need to update this?  or we leaving as 2.X.0?

taysea:
Converting to draft until functionality is released in Grommet!

.changeset/heavy-fireants-walk.md
taysea, line 5:
```suggestion
- Added `global.focus.inset` for cases like DataTable expand control which leverages `focusIndicator="inset"` internally. Supported when using grommet >=2.48.0.
```

britt6612 (approved):

PR #489, opened 2025-07-10
Fix `global.focus.shadow.blur`

taysea:
#### What does this PR do?

The intended style is for the boxShadow (used for dark mode focus indicator) to not have any blur. This fixes that. Previously it was falling back to the default `2px` blur.

Local in DS site:

BEFORE:
Screenshot 2025-07-10 at 4 11 06 PM

AFTER:
Screenshot 2025-07-10 at 4 11 00 PM

### DIFF
--- a/.changeset/nine-emus-change.md
+++ b/.changeset/nine-emus-change.md
@@ -2,4 +2,4 @@
 "grommet-theme-hpe": patch
 ---
 
-- Fixed `global.focus.blur` to be intended value of `0px` instead of `2px`.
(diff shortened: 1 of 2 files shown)


.changeset/nine-emus-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"grommet-theme-hpe": patch
+---
+
+- Fixed `global.focus.blur` to be intended value of `0px` instead of `2px`.
```
taysea, line 5:
```suggestion
- Fixed `global.focus.shadow.blur` to be intended value of `0px` instead of `2px`.
```

PR #490, opened 2025-07-21
Enhance theme-driven "bold" weight

### LINKED ISSUE #481: Remap "bold" to weight 500 in theme

taysea:
#### What does this PR do?
It is beneficial to drive what the font-weight string "bold" should resolve to via the theme. A few major versions ago, we instructed that "bold" treatment should really now use "medium" to evolve with HPE's brand, however leaving it up to individual teams to implement this poses additional remapping work and room for divergence.

This PR changes the behavior of Text / Heading such that:
- If the caller passes `weight="bold"`, it will resolve to whatever we define "bold" to mean in the theme (at the current state, 500)
- If the caller passes weight as a string/number (for example, `weight={800}` or `weight="800"`), we will respect the specific weight they passed.

This allows the concept of "bold" to be theme-driven, while still allowing the caller to override via the specific numeric string if a true bold is needed for a use case.

Locally in DS site

Screenshot 2025-07-21 at 4 30 19 PM

Closes #481

### DIFF
--- a/.changeset/hungry-donuts-crash.md
+++ b/.changeset/hungry-donuts-crash.md
@@ -1,5 +1,5 @@
 ---
-"grommet-theme-hpe": patch
+"grommet-theme-hpe": major
 ---
 
 - Updated Text and Heading such that passing `weight="bold"` will resolve to theme-defined weight of 500. This enables the concept of "bold" to be theme-driven. For use cases that need the font-face's true bold, numeric font-weights 800 can be used and won't be overridden. 
--- a/.changeset/hungry-donuts-crash.md
+++ b/.changeset/hungry-donuts-crash.md
@@ -1,0 +1,5 @@
+---
+"grommet-theme-hpe": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/hungry-donuts-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"grommet-theme-hpe": patch
```
taysea, line 2:
```suggestion
"grommet-theme-hpe": major
```

src/js/themes/hpe.js
```diff
@@ -379,11 +379,17 @@ const buildTheme = (tokens, flags) => {
       large.hpe.text?.[textSize]?.fontWeight || fallback.weight;
   });
 
-  textTheme.extend = ({ size: textSize, weight }) =>
-    !weight ? `font-weight: ${fontWeights[textSize]};` : '';
+  textTheme.extend = ({ size: textSize, weight }) => {
+    if (!weight) return `font-weight: ${fontWeights[textSize]};`;
+    if (weight === 'bold') return `font-weight: 500;`;
```
MikeKingdom, line 384: I think it would be better to have this mapping from `bold` to `500` in the theme rather than hardcoded in a per-component way. Then this logic could actually be in base.js

PR #495, opened 2025-08-13
Theme obj updates

taysea:
#### What does this PR do?

This updates t-shirt sizes on components that have existing theme objects available (separate from work being handled by new t-shirt theme objects).


src/js/themes/hpe.js
```diff
@@ -1824,14 +1825,34 @@ const buildTheme = (tokens, flags) => {
         },
       },
     },
+    nameValueList: {
```
jcfilben, line 1828: Should we also include a re-map for `nameValueList.name.width` from 'small' to 'xsmall'?

src/js/themes/hpe.js
```diff
@@ -1824,14 +1825,34 @@ const buildTheme = (tokens, flags) => {
...
+      },
+    },
     nameValuePair: {
       name: { color: 'text-strong', weight: global.hpe.fontWeight.medium },
     },
     notification: {
       close: { icon: Close },
-      container: { round: 'xsmall' },
+      container: {
+        round: 'xsmall',
```
jcfilben, line 1848: Does round stay the same (no re-mapping)?

src/js/themes/hpe.js
```diff
@@ -1824,14 +1825,34 @@ const buildTheme = (tokens, flags) => {
...
+            column: 'xlarge',
+            row: 'medium',
+          },
+        },
+      },
+    },
     nameValuePair: {
       name: { color: 'text-strong', weight: global.hpe.fontWeight.medium },
     },
     notification: {
```
jcfilben, line 1845: Looks like we are missing re-mapping `notification.gap` along with some other notification props

src/js/themes/hpe.js
```diff
@@ -1938,29 +1959,33 @@ const buildTheme = (tokens, flags) => {
       wide: {
         width: {
           min: '336px', // 336 + 24 (margin) + 24 (margin) = 384 (e.g. 'medium')
-          max: 'xxlarge',
```
jcfilben, line 1941: It looks like 'xlarge' re-maps to '3xlarge'. Does 'xxlarge' also re-map to '3xlarge'?

src/js/themes/hpe.js
```diff
@@ -1938,29 +1959,33 @@ const buildTheme = (tokens, flags) => {
...
           min: '336px', // 336 + 24 (margin) + 24 (margin) = 384 (e.g. 'medium')
           max: '100%',
         },
-        xsmall: { pad: { horizontal: 'large' } },
-        xlarge: { pad: { horizontal: 'large' } },
+        xsmall: { pad: { horizontal: 'xlarge' } },
+        xlarge: { pad: { horizontal: 'xlarge' } },
       },
     },
     pageHeader: {
```
jcfilben, line 1984: Looks like we didn't re-map `pageHeader.responsive.actions.pad`

src/js/themes/hpe.js
```diff
@@ -1974,7 +1999,16 @@ const buildTheme = (tokens, flags) => {
...
+        gap: { row: '3xsmall', column: 'medium' },
+      },
+      small: {
+        gap: { row: '3xsmall', column: 'large' },
+      },
+      medium: {
+        gap: { row: '3xsmall', column: 'medium' },
+      },
+      large: {
+        gap: { row: '3xsmall', column: 'large' },
```
jcfilben, line 2011: Looks like these have a pad defined in base.js in addition to a gap

src/js/themes/hpe.js
```diff
@@ -1984,7 +2018,7 @@ const buildTheme = (tokens, flags) => {
         ],
         columns: [['medium', 'large'], 'auto'],
         rows: ['auto', 'auto', 'auto'],
-        gap: { row: 'xsmall', column: 'large' },
+        gap: { row: '3xsmall', column: 'large' },
```
jcfilben, line 2021: Should this be 
```suggestion
        gap: { row: '3xsmall', column: 'xlarge' },
```

src/js/themes/hpe.js
```diff
@@ -2206,6 +2244,7 @@ const buildTheme = (tokens, flags) => {
       clear: {
         container: {
           pad: components.hpe.select.default.medium.drop.paddingX,
+          round: 'xsmall',
```
jcfilben, line 2247: I don't see this defined in base.js

src/js/themes/hpe.js
```diff
@@ -1301,10 +1301,11 @@ const buildTheme = (tokens, flags) => {
         // border: undefined,
         // pad: undefined,
       },
-      groupEnd: { border: { side: 'bottom', size: 'xsmall' } },
+      groupEnd: { border: { side: 'bottom', size: '3xsmall' } },
       header: {
         border: { side: 'bottom' },
         color: components.hpe.headerCell.default.rest.textColor,
+        pad: { vertical: '3xsmall' },
```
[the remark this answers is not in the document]
taysea: Hmm -- need to investigate this one.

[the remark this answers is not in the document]
taysea: We have gap handled already and I've added units.margin now -- will need to double check on the pad though in terms of why I added that.

src/js/themes/hpe.js
taysea: Done!

src/js/themes/hpe.js
taysea: Whoops the mapping table was off for that. I just fixed it. But this remapping xxlarge --> 3xlarge is correct.

src/js/themes/hpe.js
taysea: Handled adding the `pad` above ^^ was in a different theme obj than here but I was missing it.

src/js/themes/hpe.js
taysea: You're right -- with the latest from select clear styling we don't even need this at all actually.

src/js/themes/hpe.js
```diff
@@ -1984,7 +2107,16 @@ const buildTheme = (tokens, flags) => {
         ],
         columns: [['small', 'flex'], 'auto'],
         rows: ['auto', 'auto', 'auto'],
-        gap: { row: 'xsmall', column: 'medium' },
+        gap: { row: '3xsmall', column: 'medium' },
+      },
+      small: {
+        gap: { row: '3xsmall', column: 'large' },
```
jcfilben, line 2113: Should column be 'xlarge' here? Looks like it is 'large' in base.js
```suggestion
        gap: { row: '3xsmall', column: 'xlarge' },
```

src/js/themes/hpe.js
```diff
@@ -1984,7 +2107,16 @@ const buildTheme = (tokens, flags) => {
...
+      },
+      medium: {
+        gap: { row: '3xsmall', column: 'medium' },
+      },
+      large: {
+        gap: { row: '3xsmall', column: 'large' },
```
jcfilben, line 2119: Should column be 'xlarge' here? Looks like it is 'large' in base.js
```suggestion
        gap: { row: '3xsmall', column: 'xlarge' },
```

src/js/themes/hpe.js
```diff
@@ -2215,6 +2350,7 @@ const buildTheme = (tokens, flags) => {
     select: {
       clear: {
         container: {
+          background: 'transparent',
```
jcfilben, line 2353: Is this change intentional?

taysea: Yeah intentional -- when I was testing locally, the base.js grey background was still coming through on the container, but we don't want that. (related to the select clear container PR that went into NEXT)

jcfilben: Cool, thanks for confirming

jcfilben (approved):
Looks good!

PR #499, opened 2025-08-14
Add data filter theme values

Sulaymon333:
#### What does this PR do?
Add data filter theme values

Yes

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -1259,6 +1259,15 @@ const buildTheme = (tokens, flags) => {
       container: { cssGap: true, gap: 'small', margin: 'none' },
     },
     data: { button: { kind: 'toolbar' } },
+    dataFilter: {
+      rangeSelector: {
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -1259,6 +1259,15 @@ const buildTheme = (tokens, flags) => {
       container: { cssGap: true, gap: 'small', margin: 'none' },
     },
     data: { button: { kind: 'toolbar' } },
+    dataFilter: {
+      rangeSelector: {
+        size: 'full',
+        round: 'xsmall',
```
taysea, line 1265:
```suggestion
        round: 'medium',
```
I think "medium" is correct looking at the values here for radius? https://github.com/grommet/hpe-design-system/issues/4487#issuecomment-2576435972

taysea: was "small", so now should be "medium"?

PR #534, opened 2025-09-19
Setup theme objs for hpe-icons

jcfilben:
#### What does this PR do?
Prep theme for hpe-icons migration


jcfilben:
Marking as a draft until hpe-icons is ready

PR #543, opened 2025-10-03
[Hpe pop] - Added additional hpePop theme values from v2 alpha branch

halocline:
To review: Compare the hpePop theme values in this PR to those values which reside on https://github.com/grommet/grommet-theme-hpe/blob/v2-alpha/src/js/themes/hpePop.js

### DIFF
--- a/src/js/themes/hpePop.js
+++ b/src/js/themes/hpePop.js
@@ -1,3 +1,4 @@
+import { light, dark } from 'hpe-design-tokens/grommet';
 import { hpe } from './hpe';
 
 export const isObject = (item) =>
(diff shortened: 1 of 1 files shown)


halocline:
> The bolding of the headings looks a bit strong, just want to confirm that this is the correct behavior
> 

The weight is defined as 500 in Web assets: https://www.figma.com/design/WyETRsEuCm1Hd40WuLysdU/Landmark-Web-Design-System?node-id=86-237&t=NuJzZ4JYo6rLYbpA-4

jcfilben (approved):

Issue #554, opened 2025-10-27
[Bug] "info" Notification using incorrect icon


halocline:
**Expected**
"info" Notification should use `<StatusInfo />` from `@hpe-design/icons-grommet`

**Actual**
Currently using `<CircleInformation />` from `grommet-icons`

- [ ] Replace info icon used
- [ ] Check if similar adjustments need to be made for other notification statuses.

britt6612:
closing since the PR was merged https://github.com/grommet/grommet-theme-hpe/pull/557

PR #557, opened 2025-10-29
update places that grommet had icons defined to use the new hpe icons

britt6612:
#### What does this PR do?
This PR updates the area's in which Grommet was setting icons in the base theme to icons from `grommet-icons`
locally 
There were icons defined in the Grommet `base.js` file that are pointing to `grommet-icons` we need to update any reference to those icons to the new hpe-icons set
closes #https://github.com/grommet/grommet-theme-hpe/issues/554


jcfilben (approved):
Looks good!

src/js/themes/hpe.js
```diff
@@ -2270,6 +2294,7 @@ const buildTheme = (tokens, flags) => {
         },
       },
       info: {
+        icon: Info,
```
halocline: I believe this should be `StatusInfo`. See https://github.com/grommet/grommet-theme-hpe/wiki/Migrating-from-HPE-Theme-7.x-to-8.x#status-backgrounds

```suggestion
        icon: StatusInfo,
```

PR #564, opened 2025-12-02
Add theme support for thumbs and star rating components

Sulaymon333:
#### What does this PR do?
- Add theme support for thumbs and star rating components

closes https://github.com/grommet/grommet/issues/7840

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -1866,12 +1866,6 @@ const buildTheme = (tokens, flags) => {
                     .borderColor
             }; `,
         },
-        icons: {
-          like: Like,
-          likeSelected: LikeFill,
-          dislike: Dislike,
-          dislikeSelected: DislikeFill,
-        },
       },
       starRating: {
         container: {
@@ -1927,10 +1921,6 @@ const buildTheme = (tokens, flags) => {
         size: 'xsmall',
         color: components.hpe.formField.default.help.rest.color,
         margin: 'none',
-      },
-      icons: {
-        selected: StarFill,
-        unselected: Star,
       },
       info: {
         size: 'xsmall',


britt6612:
@Sulaymon333 these changes look good but want to point something out that I saw, and if you can make the changes.

so here 
https://github.com/grommet/grommet-theme-hpe/blob/master/src/js/themes/hpe.js#L1869
can you delete this block its not actually going in here 
```
        icons: {
          like: Like,
          likeSelected: LikeFill,
          dislike: Dislike,
          dislikeSelected: DislikeFill,
        },
```

also here as well looks like the Star Icon were being used but thats not an actual theme value
https://github.com/grommet/grommet-theme-hpe/blob/master/src/js/themes/hpe.js#L1931
if you can delete this block of code as well

```
      icons: {
        selected: StarFill,
        unselected: Star,
      },
```

after these changes everything else looks good!

Sulaymon333:
@britt6612 thank you for the feedback, I have removed the codeblocks as suggested via this [commit](https://github.com/grommet/grommet-theme-hpe/pull/564/commits/da4222d439f5317bc3f2eed9f79f1b58b9d7e285). Although, I did not notice any difference. The icons sizes are still 20px for  `size="xlarge"`

src/js/themes/hpe.js
```diff
@@ -3114,6 +3112,23 @@ const buildTheme = (tokens, flags) => {
     thumbsRating: {
       like: { color: 'background-selected-primary-strong' },
```
britt6612: if we are declaring `color` below in the icons block do we need these lines? @Sulaymon333

britt6612:
> > @jcfilben the whole icon needs to be the 24px correct?
> 
> The click target needs to be 24x24

in this case this would pass then how it is now. So I believe this PR is good!

Screenshot 2025-12-05 at 10 16 57 AM

src/js/themes/hpe.js
```diff
@@ -2881,6 +2871,14 @@ const buildTheme = (tokens, flags) => {
     },
     starRating: {
       color: 'background-selected-primary-strong',
```
britt6612: @Sulaymon333 you can test but I dont think you need this color in both here as well as defined in the icons

jcfilben:
@Sulaymon333 The click target should be a minimum of 24x24 pixels both when the ThumbsRating/StarRating is within a FormField and when it is outside of a FormField.

When the ThumbsRating/StarRating is in a FormField it gets extra padding around it which makes the click target bigger. Maybe we can explore if adding padding regardless of whether the Thumbs/StarRating is in a FormField. To do this we would likely need to removing the padding that gets added on FormField for these components so that we don't get a double pad. Could potentially leverage the theme.formfield.starRating.container.extend prop. You may just want to run this direction by designer too to make sure they are comfortable with that change.

Sulaymon333:
> @Sulaymon333 The click target should be a minimum of 24x24 pixels both when the ThumbsRating/StarRating is within a FormField and when it is outside of a FormField.
> 
> When the ThumbsRating/StarRating is in a FormField it gets extra padding around it which makes the click target bigger. Maybe we can explore if adding padding regardless of whether the Thumbs/StarRating is in a FormField. To do this we would likely need to removing the padding that gets added on FormField for these components so that we don't get a double pad. Could potentially leverage the theme.formfield.starRating.container.extend prop. You may just want to run this direction by designer too to make sure they are comfortable with that change.

@grommet/design please what your input on this. this will help guide the next direction.

PR #565, opened 2025-12-10
Set meter background to background-contrast

jcfilben:
#### What does this PR do?
Set theme.meter.background to `background-contrast` instead of falling back to the default grommet value.

Backwards compatible

### DIFF
--- a/src/js/themes/hpe.js
+++ b/src/js/themes/hpe.js
@@ -2167,6 +2167,7 @@ const buildTheme = (tokens, flags) => {
       },
     },
     meter: {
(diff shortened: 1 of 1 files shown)


src/js/themes/hpe.js
```diff
@@ -2167,6 +2167,7 @@ const buildTheme = (tokens, flags) => {
       },
     },
     meter: {
+      background: 'background-contrast',
```
britt6612: was there reasoning behind this color specifically?

britt6612 (approved):

PR #568, opened 2025-12-12
Change StarRating and ThumbsRating icons to use hpe set

jcfilben:
#### What does this PR do?
Removes incorrect theme code where we were attempting to set the icons within the FormField object of the theme. Adds icon assignment to the StarRating and ThumbsRating theme objects

Backwards compatible


britt6612 (approved):
LGTM

PR #577, opened 2026-01-08
move @hpe-design/icons-grommet from peerDependencies to dependencies

britt6612:
#### What does this PR do?
Moves @hpe-design/icons-grommet out of peerDependencies and into dependencies in [package.json](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). This ensures the icon package is installed when grommet-theme-hpe is installed and avoids relying on consuming projects to provide the package.

@hpe-design/icons-grommet was previously declared as a peer dependency, which requires consumers to install it explicitly. Making it a regular dependency simplifies installation for consumers of the HPE theme and reduces the chance of missing icon package errors.
backward compatible
Potential impact: Consumers that intentionally deduplicated or pinned a different version of @hpe-design/icons-grommet might observe a different dependency tree. In typical usage, the change is non-breaking it simply ensures the icons package is present when the theme is installed.

yes

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "@babel/preset-env": "^7.24.4",
     "@babel/preset-react": "^7.24.1",
     "@changesets/cli": "^2.28.1",
-    "@hpe-design/icons-grommet": "^1.0.1",
     "babel-loader": "^9.1.3",
     "cross-env": "^7.0.3",
     "del": "^6.1.1",


package.json
```diff
@@ -20,12 +20,12 @@
     "url": "https://github.com/grommet/grommet-theme-hpe.git"
   },
   "peerDependencies": {
-    "@hpe-design/icons-grommet": "^1.0.1",
     "grommet": "^2.50.0",
     "react": "^16.13.1 || ^17.0.1 || ^18.0.0 || ^19.0.0",
     "styled-components": "^6.0.0"
   },
   "dependencies": {
+    "@hpe-design/icons-grommet": "^1.0.1",
```
jcfilben: Wondering if we should go ahead and bump this version to 1.1.0 and use the new Element icon?

```suggestion
    "@hpe-design/icons-grommet": "^1.1.0",
```

britt6612: ah yes good call I would agree

package.json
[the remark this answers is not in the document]
britt6612: yes forgot it was still there good call

jcfilben (approved):

PR #630, opened 2026-09-09
Automate grommet theme hpe release

britt6612:


knowledge/capabilities/grommet-theme-hpe-publishing.md
```diff
@@ -0,0 +1,151 @@
+# grommet-theme-hpe Publishing Capability
```
halocline, line 1: I don't think we need to introduce `knowledge/capabilities` pattern in this repository. It is useful in the `hpe-design-system` repo because we are building that as an artifact for use by consumers outside of the repo's context. The release capability is specific to the repo, so it could live as a skill under .github, but chose to keep all capabilities consolidated in the monorepo for clarity.

In this case, since this is specific to this repo this should probably exist under `.github/skills/publish-grommet-theme-hpe.SKILL.md`.

britt6612: yes your are right I will update when I come back to this

PR #634, opened 2026-09-17
Fix Text & Icon strong colors

halocline:
[Deploy preview](https://deploy-preview-634--gregarious-clafoutis-525547.netlify.app/?path=/story/theme-colors--background)

#### What does this PR do?

Fixes `text-onStrong`

Storybook & updated snapshot tests

Introduction of `background-accent-{color}-strong` colors exposed that `text-onStrong` and `icon-onStrong` colors needed their light/dark mode values swapped. They were not included with the previous `onPrimaryStrong` and `onSelectedPrimaryStrong` work.

Note: `background-accent-cyan-strong` doesn't seem to meet 4.5:1 contrast in dark mode.

Screenshot 2026-09-17 at 4 40 31 PM

Screenshot 2026-09-17 at 4 37 08 PM

Screenshot 2026-09-17 at 4 38 18 PM


oliverHPE (approved):
Storybook looks correct for Button component and `background` colors (besides 
`text.Onstrong` and `icon.Onstrong` on background-accent-cyan-strong which is to be expected).

britt6612:
Looks good I had a question should we remove Cyan until we figure out the cause of the issue since its not passing accessibility? https://github.com/grommet/hpe-design-system/issues/6589

halocline:
> Looks good I had a question should we remove Cyan until we figure out the cause of the issue since its not passing accessibility? [grommet/hpe-design-system#6589](https://github.com/grommet/hpe-design-system/issues/6589)

I don't think we can remove. The color has already been released.

I think the only thing we can do is treat cyan as a one-off and handle differently for the time being -- until an approach to Grommet is figured out. Likely need a flag in the Grommet theme object to opt-out of the "background is dark" check.

britt6612:
> > Looks good I had a question should we remove Cyan until we figure out the cause of the issue since its not passing accessibility? [grommet/hpe-design-system#6589](https://github.com/grommet/hpe-design-system/issues/6589)
> 
> I don't think we can remove. The color has already been released.
> 
> I think the only thing we can do is treat cyan as a one-off and handle differently for the time being -- until an approach to Grommet is figured out. Likely need a flag in the Grommet theme object to opt-out of the "background is dark" check.

oh right yes the color was already released.. Everything else looks good so happy if you want to merge. We could do something locally maybe specific for the Cyan color as a one-off but also not sure if we want to do that might be fragile as you mentioned in the ticket.

MikeKingdom (approved):
Looks good and nice stories. I think the fix for cyan probably is an improved colorIsDark like you basically noted. I added a comment to https://github.com/grommet/hpe-design-system/issues/6589 about potentially using a better luminance calculation.

